PR #8, opened 2016-11-08
some improvements tested on debian

MartinVerges:
Hello,

it's now possible to use a external db, fixed installation problems on debian jessie systems and improved error handling with config.php

It would be nice if you accept this pull request.

Thanks
Martin


aalaesar:
Hello Martin and thank you for the Merge Request,
I have tested your changed and they are ok. 👍 
I have just a question about the default value for `nextcloud_install_db : false`
from my perspective, it breaks the role's simplest use (case 1 in the readme) and must be specified every time the role is used for an "all in one" server (which is the main use case).
I whould recommend "true" as default.
What's your point of view ?
Regards,
Aal.

MartinVerges:
yes, default true would be better for the public version.

aalaesar:
Thank you for your answer @MartinVerges.
Let's merge !

PR #10, opened 2016-11-27
Make strong Apache TLS settings optional

jaywink:
Default to true, but allow disabling the strong TLS settings for Apache. Rationale is that many web servers will have many sites running on them. And commonly, this kind of TLS configuration will already be in place in global configuration files.


aalaesar:
Thank you for the contribution @jaywink ! 👍

PR #12, opened 2016-11-27
Allow specifying nextcloud_os_user to create a user with home folder

jaywink:
This allows then placing webroot and data dir under the created user.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -10,6 +10,10 @@ nextcloud_data_dir: "/var/ncdata"
 nextcloud_admin_name: "admin"
 # nextcloud_admin_pwd: "secret"
 nextcloud_dl_url: "https://download.nextcloud.com/server/releases"
+# Set a username here if a user with home folder should be created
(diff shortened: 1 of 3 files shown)


aalaesar:
I'm not sure I fully understand this use case.
could you elaborate ?
Regards, Aal.

jaywink:
This is probably more of how I like to organize apps on my server. I prefer them to be under home folders and their own dedicated users instead of being spread around. Though I realize this doesn't bring that much benefit in a PHP app which needs to run under the webserver user.

It's probably not that useful to this role, I just wanted to offer all the mods I made for my deployment back to your role. Happy that some of them were accepted 👍 But yeah, I'll close this one.

Thanks for your work on this role btw!

aalaesar:
I understand your point @jaywink. I do also use this method to organise my files and services on my hosts. :)
As you pointed it out, the Nextcloud app requires webserver user as owner, therefore this security isolation is lost (kind of).
However, the instance's root directory can be specified with `nextcloud_webroot` to go to a dedicated home folder.
Anyway, thank you for your help on the role.
Regards, Aal

PR #13, opened 2016-11-27
Fix conditionally including nextcloud download and install

jaywink:
If downloaded or configured, skip download or configuration respectively.


aalaesar:
Nice addition. @jaywink Thank again 👍 !

PR #15, opened 2017-01-08
Allow skipping of web and database server installation.

noplanman:
This PR adds 2 extra options to skip webserver (and PHP) installation and TLS certificate management.

Just like the database installation can be skipped, this allows for the webserver and TLS part to be skipped.

I use this role in an environment that already has DB and reverse proxy set up, so there's no need for a duplicate installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ This role requires root access, so either configure it in your inventory files, 
 
 ## Role Variables
 
-Role's variables (and their default value):
+Role's variables (and their default values):
 
 ### Installation configuration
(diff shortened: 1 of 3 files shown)


aalaesar:
Hi @noplanman , that's a good addition, thank you for the pull request.
Just before i merge it, could you add some infos on the Readme, 
thank you :-)

noplanman:
@aalaesar Have added short descriptions, is that ok?

aalaesar:
It's perfect, Thank you :+1:

PR #17, opened 2017-01-10
Non TLS webserver

### LINKED ISSUE #16: Allow usage on port 80 behind proxy
I use this role to set up Nextcloud on a server that sits behind a Nginx reverse proxy.

The proxy manages all TLS certificate stuff and basically just needs to connect to a simple port 80 webserver that is hosting the Nextcloud instance.

Would be nice to have an option to allow this setup, instead of forcing TLS.

I'm happy to PR this if you like, as I use this in my setup anyway 👍

noplanman:
What this PR contains:
- Allows the apache and nginx servers to be set up to not require TLS.
- Fixes a bug that wasn't adding all aliases to the apache configuration correctly (length-1 [here](https://github.com/aalaesar/install_nextcloud/pull/17/files#diff-7fbbb660a98bb360e8b59c996551b011L12)).
- Reload apache correctly after the configuration has changed.
- **HSTS option added!** ~I've commented out the HSTS code, as it can be dangerous if used without knowing what it does. **I'd highly suggest to add an option for this, as otherwise the configuration would always be overwriting the value!**~

@aalaesar Please check it out and then I'll think of how to add an appropriate entry to the reamde.

Fix #16

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,13 @@ Force http to https.
 ```YAML
 nextcloud_force_strong_apache_ssl: true
 ```
-force strong ssl configuration in the virtualhost file
+Force strong ssl configuration in the virtualhost file.
+```YAML
+nextcloud_hsts: false
+```
+Set HTTP Strict-Transport-Security header (e.g. "max-age=15768000; includeSubDomains; preload").
+
+*(Before enabling HSTS, please read into this topic first)*
 ```YAML
 nextcloud_tls_cert_method: "self-signed"
 ```
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -37,6 +37,7 @@ nextcloud_tls_enforce: true
 nextcloud_force_strong_apache_ssl: true
 nextcloud_tls_cert_method: "self-signed"
 nextcloud_tls_dhparam: "/etc/ssl/dhparam.pem"
+nextcloud_hsts: false
 # nextcloud_tls_cert_method: "self-signed" | "signed" | "installed"
 # nextcloud_tls_cert: /path/to/cert
 # nextcloud_tls_cert_key: /path/to/cert/key
--- a/tasks/http_apache.yml
+++ b/tasks/http_apache.yml
@@ -56,11 +56,11 @@
   template:
     dest: /etc/apache2/sites-available/nc_{{ nextcloud_instance_name }}.conf
     src: templates/apache_nc.j2
+  notify: reload apache
 
 - name: "[APACHE] -  Enable Nextcloud site in apache conf"
   file:
     path: /etc/apache2/sites-enabled/nc_{{ nextcloud_instance_name }}.conf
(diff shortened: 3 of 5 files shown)


aalaesar:
Hi @noplanman , everything looks fine from my point of view.

noplanman:
Shall I add an option for the HSTS in this PR while I'm at it?

aalaesar:
It seems I know little about the risks of HSTS, can you explain me more please ?

noplanman:
Sure! I'd recommend you search online for more resources to deepen your knowledge, but I can give you a nutshell version here :blush:

[HSTS](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) (HTTP Strict Transport Security) forces the use of HTTPS. This is managed by the browser, which reads the header and acts on it. Basically, the HSTS header tells a browser that the site MUST be accessed through HTTPS (HTTP access will give you an error).

In general practice this is brilliant, to make sure that users are connected to your website securely.

The problem lies in the mistaken use of it.
- The `max-age` flag tells the browser how long the site should be remembered as HTTPS-only. So if your site goes offline or you move the domain to a new server, the browser will still remember that only HTTPS is allowed! So the new server must support it too.
- The `includeSubDomains` (as the name says), tells the browser that ALL subdomains must also be accessed via HTTPS. This makes it impossible to have any subdomain with HTTP only (for testing something or simple APIs that don't require authentication, etc.).
- Adding the `preload` flag tells browsers to include your HSTS request into the browser itself, making even the first request to your website secure, going through HTTPS. For this, you must add your domain to the [HSTS preload list](https://hstspreload.appspot.com/).

So in general, it's a great feature, but if somebody adds it to test and puts all the above flags and a large `max-time`, they are forced to serve content via HTTPS, or their domain becomes unusable!

Ok, that should be the nutshell version of it, you can find looots of info online about this.
I can highly recommend using it, for obvious reasons, but the user really should know what's going on and the risks.

Also interesting, but even more dangerous, is [HPKP](https://en.wikipedia.org/wiki/HTTP_Public_Key_Pinning) (HTTP Public Key Pinning)

Right, I'll go ahead and add an option for this, making the default `false`, forcing the user to actively add a `true` flag. Ok?

aalaesar:
Thank you for the nutshell explanation, I didn't known about the misuse complications.
I just checked in the recommended Nextcloud installation and It is not activated in their recommandations.
So **Ok** for the option `false `by default.

noplanman:
Have added the `nextcloud_hsts` option, `false` by default.

To set it, user can simply add the header value itself, e.g.:
`nextcloud_hsts: "max-age=15768000; includeSubDomains; preload"`

This makes sure the user must know what to put in there, hoping they've read up and learned about it :smiley:

aalaesar:
Awesome Work ! :+1: 
Thanks @noplanman .

PR #20, opened 2017-01-11
Fix occ command

### LINKED ISSUE #7: occ installation do not support password beginning with "-"
When a password starts with a "-" the occ command sees it as an option call, not as a string
Throws the error: 
`The "--admin-pass" option requires a value.

noplanman:
Fix the occ command to allow any characters to be passed as parameters.

To pass quotes as a password, the variable must be defined properly:
`"i'm a password"` would be `nextcloud_admin_pwd: '\""I\''m a password\”"'`

The formatting may be a bit weird, but luckily it only needs to be set once 😉 
(if you have a password with quotes that is...)

Fix #7

### DIFF
--- a/tasks/nc_installation.yml
+++ b/tasks/nc_installation.yml
@@ -40,15 +40,15 @@
   - name: "[NC] - Run occ installation command"
     become_user: "{{ websrv_user }}"
     command: >
-        php occ  maintenance:install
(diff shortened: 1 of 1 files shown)


aalaesar:
Damned ! It was that simple...
Awesome @noplanman , thank you for the PR

PR #23, opened 2017-01-17
Prefixing

noplanman:
This PR prefixes default variables and ones that get registered during the role.

To prevent any clashes with other roles that might be using the same (partly generic) variable names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,15 +174,15 @@ Defines various method for retrieving a TLS certificate.
 
 ### System configuration
 ```YAML
-websrv_user: "www-data"
+nextcloud_websrv_user: "www-data"
 ```
 system user for the http server
 ```YAML
-websrv_group: "www-data"
+nextcloud_websrv_group: "www-data"
 ```
 system group for the http server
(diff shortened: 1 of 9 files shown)


aalaesar:
I find most the variables names even more generic without the nextcloud|nc prefix and therefore for prone to clash
Is there any story behind this change ? Any production issue ?

noplanman:
No production issue as of yet.

This is more a preventive move, in case some other role does the same thing and has variables called `db_install` for example.
Best practice (as far as I have seen/learnt from other roles) is to prefix everything to prevent any issues at some point in the future.

I didn't fully understand your comment about the generic variable names. Generic is bad in this situation, as roles can overwrite each others values.

aalaesar:
> I didn't fully understand your comment about the generic variable names. Generic is bad in this situation, as roles can overwrite each others values.

Never mind ... coffee didn't kicked in this morning... I completely misread and inverted the PR changes (nextcloud_something => something).
That's a good PR :+1:

PR #25, opened 2017-01-19
Make travis test rolename agnostic.

noplanman:
This is so that the test works for users that have renamed the role.


noplanman:
Same as the upgrade role, is this good for merge?

noplanman:
Ok, this is done.

I really need this to be able to run tests on my branch 👍

PR #26, opened 2017-01-30
Rework version selection #18

### LINKED ISSUE #18: Confusion with daily versions
Looking at how the Nextcloud version can be specified, something's a bit off.

The main issue is when choosing the daily channel, namely this part:
`daily: "nextcloud-{{nextcloud_branch}}-daily-{{nextcloud_version}}.zip"`

Looking into the [daily folder](https://download.nextcloud.com/server/daily/) on the download server, this format doesn't exist.

For daily stable releases it should be something like:
`latest-stable{{ nextcloud_version_major }}.zip`

And for bleeding edge daily is should be:
`nextcloud-{{ nextcloud_version_major }}-daily-{{ ansible_date_time.date }}.zip`

I've been working on a PR for this, but I'm not sure how best to tackle it. Main issue is the date for the bleeding edge daily archives, as it's possible that the date is 1 day behind (depending on timezone), resulting in an invalid URL. For that, I was thinking of a fallback system that would go back 1 day in case the archive doesn't exist.
Other option would be to put the date in a variable, allowing to choose an specific daily version, but not sure how necessary that would be. (I guess it could be an optional thing, default being today's date, falling back to yesterday's).

What do you think?

aalaesar:
Fix #18 
- Better variables name
- Default download to latest.zip
- Can technically still download anything available in the nextcloud repo
- Added variable to override the generated download link
- Added variable for changing the file format ( zip, bz2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ nextcloud_version_major: 10
 ```YAML
 nextcloud_version_channel: "releases"
 nextcloud_get_latest: false
-nextcloud_full_version: 10.0.3
+nextcloud_full_version: "10.0.3"
 ```
 - Get the nextcloud 11.0.1 prerelease 1:
 ```YAML
@@ -125,13 +125,13 @@ nextcloud_get_latest: true
 ```YAML
 nextcloud_version_channel: "daily"
 nextcloud_get_latest: true
-nextcloud_version_major: stable10
+nextcloud_version_major: "stable10"
 ```
 - Get the daily for master at january 1rst 2017:
 ```YAML
 nextcloud_version_channel: "daily"
 nextcloud_get_latest: false
-nextcloud_version_full: master
+nextcloud_version_major: "master"
 nextcloud_version_special: "2017-01-01"
 ```
 ### Main configuration
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ nextcloud_archive_format: "zip" # zip | tar.bz2
 Choose between the 2 archive formats available in the repository.
 
 ```YAML
-# nextcloud_archive_url:
+# nextcloud_full_url:
 ```
 _If you don't like rules..._
 Specify directly a full URL to the archive. The role will skip the url generation and download the archive
--- a/README.md
+++ b/README.md
@@ -21,35 +21,74 @@ This role requires root access, so either configure it in your inventory files, 
 > playbook.yml:
 ```YAML
 - hosts: dnsserver
+  become: yes
   roles:
     - role: aalaesar.install_nextcloud
-      become: yes
 ```
 
 ## Role Variables
 
 Role's variables (and their default values):
 
-### Installation configuration
-> Source location will be calculated following channel, version and branch values.
+### Choose the version
+
+An URL will be generated following naming rules used in the nextcloud repository
+_Not following this rules correctly may make the role unable to download nextcloud._
+
+#### Repository naming rules:
+Some variables changes depending on the channel used and if get_latest is true.
+This table summarize the possible cases.
+
+|channel|latest|major&latest|major|full|special|
+|---|---|---|---|---|---|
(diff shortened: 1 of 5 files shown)


aalaesar:
### TODO
1. Add support for bz2 files
2. Complete readme.

aalaesar:
Hi @noplanman,
I'm about to merge.
Any comment on the reworked variables?

noplanman:
@aalaesar 
Aaargh, sorry, didn't get a chance to check this out yet as I'm abroad at the moment.
Will have a look now!

README.md
```diff
@@ -21,35 +21,74 @@ This role requires root access, so either configure it in your inventory files,
...
-      become: yes
 ```
 
 ## Role Variables
 
 Role's variables (and their default values):
 
-### Installation configuration
-> Source location will be calculated following channel, version and branch values.
+### Choose the version
```
noplanman: This is quite difficult to understand, maybe you could add various examples to elaborate for different scenarios?
Kind of like the Ansible documentation does it.

tasks/nc_download.yml
```diff
@@ -1,6 +1,21 @@
...
+- name: "[NC-DL] - bunzip2 is installed"
+  package: name=bzip2 state=installed
+  when: nextcloud_archive_format == "tar.bz2"
+
+- name: "[NC-DL] - regenerate the download URL for *latest*."
+  set_fact:
+    nextcloud_full_url: "{{nextcloud_repository}}/{{nextcloud_version_channel}}/{{[nextcloud_dl_file_name.latest, nextcloud_archive_format]|join('.')}}"
+  when: nextcloud_get_latest and (nextcloud_version_channel != 'prereleases') and (nextcloud_archive_url is not defined)
+
+- name: "[NC-DL] - use the provided URL instead."
```
noplanman: I'd move this one before the other one above and then check if `nextcloud_full_url` is set, to skip all the current `when` checks.

Also, why not only use `nextcloud_archive_url` **or** `nextcloud_full_url`? That would save 1 variable name and prevent any ambiguity/confusion between the two.

vars/main.yml
```diff
@@ -1,14 +1,8 @@
...
-  prereleases: "prereleases"
-  daily: "daily"
-  latest: "daily"
-nextcloud_dl_file:
-  releases: "nextcloud-{{nextcloud_version}}.zip"
-  prereleases: "nextcloud-{{nextcloud_version}}.zip"
-  daily: "nextcloud-{{nextcloud_branch}}-daily-{{nextcloud_version}}.zip"
-  latest: "latest-{{nextcloud_branch}}.zip"
-nextcloud_full_url: "{{nextcloud_repository}}/{{nextcloud_dl_path[nextcloud_channel]}}/{{nextcloud_dl_file[nextcloud_channel]}}"
+nextcloud_dl_file_name:
```
noplanman: I assume you've tested all the different possibilities here? :grin:

README.md
aalaesar: Yeah, I also thought about this, this afternoon.
I'll do simple examples just below the vars declarations, then add some variation to the final cases..

tasks/nc_download.yml
aalaesar: yes, good point!
nextcloud_full_url is defined in the vars folder therefore, always defined.
But maybe i'll simply remove it from vars and put a set_fact task here instead.
I'll use `nextcloud_archive_url`

vars/main.yml
aalaesar: most of them in a dedicated playbook yes. 😄

aalaesar:
Hi,
@noplanman 
I added 2 commits following your reviews.

noplanman:
Nice, great examples in the readme 👍 

Looks good to merge!

aalaesar:
Merged yesterday and released 1.3.0
:+1:  Thank you very much for your contributions so far @noplanman !! :+1:

PR #29, opened 2017-06-17
remove duplicate and already set variables

pheanex:
so its more consistent with the already used way.
Also use UNIX line-endings consistently (before it was mixed CRLF/LF)


aalaesar:
Thank you @pheanex for the contribution 👍  😄

PR #30, opened 2017-06-20
Allow support for Debian Stretch

eythian:
This seems to be all that's needed to get a working install on Stretch.


aalaesar:
Indeed it work
tested 👍 Thank you for the contribution !

PR #32, opened 2017-08-13
Set apache log-file in configuration

pheanex:
I explicitly set the log file for the apache configuration to the acces.log and error.log to make fail2ban integration easier.
Default (depending on your distribution i guess) is other-vhosts-acces.log.
I don't know if you care, or have a different opinion here, so this is just a "take if like, decline if not"-PR.


pheanex:
Nevermind ... i changed my mind. Sorry for the confusion (default logging is also okay for fail2ban)

PR #38, opened 2018-01-24
APP config value as one string

teadur:
Make it possible to use spaces inside app conf strings


aalaesar:
Nice change ! simple and effective 👍 
Thank you.

PR #40, opened 2018-04-11
All instances of state: installed changed to state: present.

arenius:
As of Ansible 2.5 'state: installed' is deprecated and will generate warnings
when used.  Used 'state: present' instead.

The warning is:
[DEPRECATION WARNING]: State 'installed' is deprecated. Using state 'present'
instead.. This feature will be removed in version 2.9. Deprecation warnings can
 be disabled by setting deprecation_warnings=False in ansible.cfg.


aalaesar:
Thank you @arenius.
 I changed this in the  manage-bind role but I forgotten to do it here.
Regards, Aal.

PR #45, opened 2018-10-09
Added option to enable ipv6 support in nginx

belidzs:


belidzs:
hi @aalaesar! Thanks for the quick reply. I agree that the variable name might be too specific, however according to the research I've done before implementing there's no need to specifically enable ipv6 in Apache as it is [on by default](https://www.rackaid.com/blog/how-to-disable-ipv6-in-apache-server/).

On the other hand other modules (like mysql/mariadb server) might also need their config changed before we can say that the install is fully ipv6 enabled. Hence the specific variable name :)

Anyway, I've added you as a collaborator on my fork (I hope this is what you had in your mind when you said commit on the PR), so feel free to do any modification before pulling the final results!

aalaesar:
Thank you @belidzs 👍 
sorry I haven't been available lately. 
merging now.

PR #46, opened 2018-10-09
Changed nginx template to implement changes suggested by NextCloud 14 release notes

belidzs:
The implemented changes were suggested by the [release notes](https://docs.nextcloud.com/server/14/admin_manual/release_notes.html) of NextCloud 14

For background compatibility I used Jinja `if` statements to only include these changes when `nextcloud_version_14_or_higher` is set to `true`. 

Since new installs will probably go with the newest version anyway, I've set its default value to `true`.

However it would be a lot more elegant if the value of this variable would be calculated based on other `nextcloud_version_...` variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,11 +94,13 @@ Choose between the 2 archive formats available in the repository.
 # nextcloud_full_url:
 ```
 _If you don't like rules..._
-Specify directly a full URL to the archive. The role will skip the url generation and download the archive
+Specify directly a full URL to the archive. The role will skip the url generation and download the archive. **Requires nextcloud_version_major to be set along**.
 #### Examples:
 - Download your own archive:
+  (_you **must** specify the nextcloud major version along_)
 ```YAML
 nextcloud_full_url: https://h2g2.com/42/nexcloud.zip
+nextcloud_version_major: 42
 ```
 - Choose the latest release (default):
 ```YAML
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -13,7 +13,6 @@ nextcloud_get_latest: true # mandatory # specify if the latest archive should be
 # nextcloud_version_major: 10 # (9 | 10 | 11| ..) for releases |  for daily : (master | stable9 | stable10 | ...)
 # nextcloud_version_full: "10.0.3" # full version string
 # nextcloud_version_special: "" #  For prereleases: "RCn|beta" | for daily "YYYY-MM-DD"
-nextcloud_version_14_or_higher: true
 nextcloud_repository: "https://download.nextcloud.com/server" # Domain URL where to download Nextcloud.
 nextcloud_archive_format: "zip" # zip | tar.bz2
 # nextcloud_full_url: "https://h2g2.com/downloads/42/my_nexcloud.zip" # specify directly a full URL to the archive if you don't like rules.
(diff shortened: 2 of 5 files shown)


aalaesar:
Hi again @belidzs 
Thank you for the contribution

I was also privately working on this one but I wanted to find it "automatically" in order to keep the version selection system not more complicated [than it is already](https://github.com/aalaesar/install_nextcloud#repository-naming-rules).
While testing, I concluded that releases channels  and also the "special" type are not relevants to find if the version is newer than 13
Then I made this truth table  with true conditions. _imply each time, "the variable is defined and ..."_

|OR|latest|major|full|
|---|---|---|---|
|-|yes and (**major** is _null_) |>13\|master|split(".")[0] >13|

I can implement this condition as a private boolean instead of the new variable on you PR _if you allow me_.

Regards, Aal.

belidzs:
hi @aalaesar! As you can see from the commit history I tried to reverse engineer the version selection system to make this work automatically but I eventually decided against it as many edge cases would have to be handled (like nightly builds after NC14 was released, etc) and it was too much work for me at the moment.

So I really don't mind if you pick up from here either in my repo or in yours. It's totally up to you :) I've granted the neccessary write permission already. 

Thanks for your great work on this!

aalaesar:
hi @belidzs,
What were the edge cases you were thinking about ? (may spare me some time 😄  )

With the truth table the only edge case I can think of (for now) is when the user provides an **offline** zip/tar.gz package
but this could be mitigated by making `nextcloud_version_major` mandatory _in this case_

(also updated the truth table for the daily channel)

belidzs:
To be honest the rule system for selecting versions was a bit hard to understand to me and this is why I stopped "reverse engineering" it.

For example I wasn't sure how illegal combinations were handled or what to do when the `daily` or `special` variables are defined. Now I start to understand but I still feel like you are better suited for this job.

aalaesar:
> To be honest the rule system for selecting versions was a bit hard to understand to me and this is why I stopped "reverse engineering" it.

You're right, the parameters and how they work are poorly explained. I initially tried a text/table only ReadMe but It will be much easier to understand with some graphic representation. 

> For example I wasn't sure how illegal combinations were handled or what to do when the `daily` or `special` variables are defined. Now I start to understand but I still feel like you are better suited for this job.

the "heart" of the "link computing system" is in  [vars/main.yml](https://github.com/aalaesar/install_nextcloud/blob/master/vars/main.yml)
It contain one dict with 4 keys. each one representing some kind of jinja2 regex-like for the archive name.

Then in [tasks/nc_download](https://github.com/aalaesar/install_nextcloud/blob/master/tasks/nc_download.yml#L10)
A block a 2 tasks emulate an `If,  then,  else` condition  In case the user choosed _latest_ ,because the file name **is** `latest` in this case.

I'll drawn a finite state machine to explain it in the readme. It may be more understandable.

Regards,
Aal

aalaesar:
hi @belidzs 
I've made [this diagram](https://raw.githubusercontent.com/aalaesar/install_nextcloud/update/readme/files/nextcloud_choosing_version.png) in order to describe how nextcloud names its archives
do you think it make the parameters clearer ?
their is still some typos but the logic is here.

any suggestion is welcome
Regards,
Aal.

aalaesar:
hi @belidzs 
whould you like to review the 3 last commits ?
two brains are better than one :smile:

aalaesar:
Ok.
no pb.
thank you for the feed back. 😃 
Regards.
Aal.

PR #49, opened 2018-11-25
Update and rename php7.0_env.yml to php7.2_env.yml

mihailstoynov:
On ubuntu 18.04 php7.0 is php7.2


aalaesar:
Hi @mihailstoynov
Thank you for the contribution
just asking myself if it might break retro-compatibility with ubuntu 16.04 and debian 9
Also I don't see a condition for bionic. Does it work ?

mihailstoynov:
I'm not sure, easiest way is to put both defaults in.
I guess i didn't think about backwards compatibility

aalaesar:
#57 fixed retro compatibility issues with changing php version

PR #50, opened 2018-11-25
php7.2-mcrypt removed

mihailstoynov:
php7.2-mcrypt removed

php/mcrypt deprecated in 7.1, removed in 7.2
nextcloud should work fine without it:
nextcloud/server#8265


aalaesar:
#57 fixed retro compatibility issues with changing php version

PR #54, opened 2018-11-25
no browser will open unencrypted (or badly encrypted) http2 traffic

mihailstoynov:
`listen 80 http2;` makes no sense

### DIFF
--- a/templates/nginx_nc.j2
+++ b/templates/nginx_nc.j2
@@ -19,9 +19,9 @@ server {
     server_name {{ nextcloud_trusted_domain | join(' ') }};
 
 {% if not nextcloud_install_tls or not nextcloud_tls_enforce %}
-    listen 80{% if (nextcloud_computed_major_version|int) > 13 %} http2{% endif %};
+    listen 80;
 {% if nextcloud_ipv6 %}
-    listen [::]:80{% if (nextcloud_computed_major_version|int) > 13 %} http2{% endif %};
+    listen [::]:80;
 {% endif %}
 {% endif %}
 {% if nextcloud_install_tls %}


aalaesar:
That's interesting.
Could you elaborate ?
I've searched some infos about http/2 and its specs, at least, says it is designed for both encrypted and unencrypted services.
I finally found some info on the HTTP/2 Wikipedia article about [discutions on mandatory encryption ](https://en.wikipedia.org/wiki/HTTP/2#Encryption_2)

But no info about how browsers deal with this case...

mihailstoynov:
AFAIR Initially http/2 was a good excuse to require encryption. When SPDY (predecessor to http/2) was rolled out, it was decided to enforce strong encryption - forward secrecy, strong ciphers. Http/2 is just a binary protocol. it itself does not require encryption, but browsers decided they will support http/2 only over strong crypto, so they can force the world to use https.

Http2 and encryption:
https://http2.github.io/faq/#does-http2-require-encryption

Browsers still only support http2 over https:
https://caniuse.com/#feat=http2
(there isn't a single browser that will support http/2 without encryption)

Http/2 became popular when EFF and Mozilla rolled out letsencrypt - so free certificates and it was much easier to force people to use https.

aalaesar:
hi @mihailstoynov 
sorry it's been a while

I am thinking about the M2M use case:
when nextcloud is be behind a reverse proxy/loadbalancer, if the RP/LB manage front-end encryption, is there any advantage for the nextcloud server to be in html/2 without encryption ?
That's may not be a case this role make straightforward.

mihailstoynov:
Hey,
That's a difficult question.
Will a LB/RP support http2 in cleartext (http2c)?
Also a LB/RP that exposes http2 and forwards to http/https (1.1) will work just fine. I don't know if there are going to be any optimizations if the service is http2c.
Also a service without ssl behing LB/RP? Issues with man in the middle?

If I were you, I wouldn't bother supporting http2c because if I want to test it, and don't add SSL, the browser is going to reject it. That's why I forked your repo.

PR #53, opened 2018-11-25
become_user requires become:yes

mihailstoynov:
Without this, the error is:

```
OCP\\Files\\NotFoundException in /opt/nextcloud/lib/private/Files/Node/Node.php:97
Stack trace:
#0 /opt/nextcloud/lib/private/Files/Node/Node.php(216): OC\\Files\\Node\\Node->getFileInfo()
#1 /opt/nextcloud/lib/private/Files/Node/Node.php(117): OC\\Files\\Node\\Node->getPermissions()
#2 /opt/nextcloud/lib/private/Files/Node/Folder.php(155): OC\\Files\\Node\\Node->checkPermissions(4)
#3 /opt/nextcloud/lib/private/Files/AppData/AppData.php(94): OC\\Files\\Node\\Folder->newFolder('css')
#4 /opt/nextcloud/lib/private/Files/AppData/AppData.php(114): OC\\Files\\AppData\\AppData->getAppDataFolder()
#5 /opt/nextcloud/lib/private/Template/IconsCacher.php(70): OC\\Files\\AppData\\AppData->newFolder('icons')
#6 [internal function]: OC\\Template\\IconsCacher->__construct(Object(OC\\Log), Object(OC\\Files\\AppData\\Factory), Object(OC\\URLGenerator))
#7 /opt/nextcloud/lib/private/AppFramework/Utility/SimpleContainer.php(81): ReflectionClass->newInstanceArgs(Array)
#8 /opt/nextcloud/lib/private/AppFramework/Utility/SimpleContainer.php(98): OC\\AppFramework\\Utility\\SimpleContainer->buildClass(Object(ReflectionClass))
#9 /opt/nextcloud/lib/private/AppFramework/Utility/SimpleContainer.php(119): OC\\AppFramework\\Utility\\SimpleContainer->resolve('OC\\\\Template\\\\Ico...')
#10 /opt/nextcloud/lib/private/ServerContainer.php(132): OC\\AppFramework\\Utility\\SimpleContainer->query('OC\\\\Template\\\\Ico...')
#11 /opt/nextcloud/lib/private/Server.php(969): OC\\ServerContainer->query('OC\\\\Template\\\\Ico...')
#12 /opt/nextcloud/3rdparty/pimple/pimple/src/Pimple/Container.php(118): OC\\Server->OC\\{closure}(Object(OC\\Server))
[...]

PR #55, opened 2018-11-25
become_user requires become:yes

mihailstoynov:
This was fixed everywhere

I don't understand why `become: yes` is not added at all the places. Am I missing something?

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -70,6 +70,7 @@
     args:
       chdir: "{{ nextcloud_webroot }}"
     become_user: "{{ nextcloud_websrv_user }}"
+    become: yes
     changed_when: false
     register: nc_apps_list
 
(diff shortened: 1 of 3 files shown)


aalaesar:
Depends on your ansible version.
Back when I built the role,
I had a lot of trouble for running occ commands with the correct user.
for some reason it run the command line with the root user instead of www-data or whatever is its name
Had to troubleshoot and finally get something working without the `become: yes`
I'll do more test with 2.4 & 2.5 ansible versions to see if to work correctly now

aalaesar:
By the way.
is this PR make obsolete the 2 other ones ?
#52 & #53

mihailstoynov:
Yes, it does obsolete #52 and #53.

I haven't tested with previous versions. Why would someone use older versions of ansible?

aalaesar:
thank you @egmont1227 for the feedback 👍  😃 
I have so few time available to work on this since summer.

Issue #31, opened 2017-07-21
/root/.my.cnf prevents to use other mysql tools


laurentj:
db_mysql.yml creates a /root/.my.cnf, with credentials of the root user of mysql.

Is it really needed ?

This file prevents some tools, like backup-manager (executed with the linux root user), to work correctly, when it uses an other mysql user than root. 

For example, backup manager is using this command :

```
/usr/bin/mysqldump --defaults-extra-file=/root/.backup-manager_my.cnf --opt -ubackup -hlocalhost -P3306 --all-databases
```

Even if the password of the 'backup' user is into /root/.backup-manager_my.cnf,  /root/.my.cnf has priority over this file, and so it take the password stored into /root/.my.cnf. 

/root/.my.cnf should not contain any credentials.

(and I don't want to do backups with a user that have all privileges).

aalaesar:
Hello @laurentj and thank you for submitting this issue.
I apologize for my late answer, I was away.
You have a good point.
The .my.cfg is used by the mysql modules to access root privileges in order to create the nexcloud user and database.
This has been made to use easily most of database managment operation.
I'll look on this point in the next few days.
Regards, Aal.

egmont1227:
This should be fixed with PR #58

aalaesar:
should be fixed by #58

PR #58, opened 2019-01-02
#31 use debian.cnf to create root pw

egmont1227:
Hey

This fix Issue #31. It will use /etc/mysql/debian.cnf to make sure correct root pw is set and uses this password on every mysql_user call.

It also contains some reformatting.


aalaesar:
Again Thank you @egmont1227 👍 !

PR #64, opened 2019-05-31
Disable the default site of apache and nginx

santiagomr:
Hi, thanks for sharing this excellent role.

I tell you that I mounted an instance on Apache with forced HTTPS (`nextcloud_tls_enforce: true`) in which everything went fine except for one detail. If I access my instance directly with the URL (`my-nextcloud.org`), I match to the default Apache site instead of my Nextcloud. Therefore I must explicitly specify the HTTPS prefix to access the site (`https://my-nextcloud.org`)

The configuration file of the Nextcloud site is generated correctly (making a permanent redirection of HTTP traffic to HTTPS), but the default Apache site (`000-default.conf`) is still enabled and traps me first.

In my opinion the solution is simply disable the default site, which is what I propose in this PR (also for Nginx).

I can't imagine a scenario in which someone wanted to keep this site enabled in production, so I did not consider it necessary add a variable to make a conditional about this task.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,10 @@ nextcloud_websrv: "apache2"
 ```
 The http server used by nextcloud. Available values are: **apache2** or **nginx**.
 ```YAML
+nextcloud_disable_websrv_default_site: false
+```
+Disable the default site of the chosen http server. (`000-default.conf` in Apache, `default` in Nginx.)
+```YAML
 nextcloud_websrv_template: "templates/{{nextcloud_websrv}}_nc.j2"
 ```
 The jinja2 template creating the instance configuration for your webserver.
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -31,6 +31,7 @@ nextcloud_instance_name: "{{ nextcloud_trusted_domain | first }}"
 
 nextcloud_install_websrv: true
 nextcloud_websrv: "apache2" # "apache2" | "nginx"
+nextcloud_disable_websrv_default_site: false
 nextcloud_websrv_template: "templates/{{nextcloud_websrv}}_nc.j2"
 nextcloud_webroot: "/opt/nextcloud"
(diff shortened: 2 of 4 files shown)


aalaesar:
Hi @santiagomr 
thank you for your feed back !
I noted this bad behavior  today while working on improving the role on ubuntu 18.04 and debian 9.
Removing the default site for apache or nginx could be a easy workaround .... but .... :thinking: this could break the user's web site :thinking: 
It's true that the best use case is to install nexcloud from scratch and not using the server for something else
but what in the case the operator added custom configuration in the default ?
For instance, the apache's default site contain logging directive while the nextcloud server file  does not.
disabling the default may disable logging

in the case of nginx this is more straightforward, but the risk to unintentionally breaking the users web's pages configuration is noticeable.

this may need more tweaks on the template generating the configuration directives for each "hostname"  used to access the nextcloud instance.
May be it's bad practice, i'll look into it

Conclusion:
> need more impact study and search of a risk free fix

Regards,
Aal

santiagomr:
Hello again, you are right that disabling the default site can cause many problems in certain environments and scenarios.

Maybe the best solution is include this characteristic in conditional to a variable (default set to `false`), as proposed in [this commit](https://github.com/aalaesar/install_nextcloud/pull/64/commits/4808af6605f0832ec657316748ddee364ece2612).

P.S. I already made a rebase on top of `master` to update these months of progress.

aalaesar:
hello @santiagomr 
thank you for the PR
sorry for the late answer.

santiagomr:
Thanks to you. Could you reimport the role in Ansible Galaxy to have these changes available?

aalaesar:
> Thanks to you. Could you reimport the role in Ansible Galaxy to have these changes available?

A new maintenance release has been made to trigger imports

Issue #65, opened 2019-06-03
Detail generating conf files with IPv6 addresses


santiagomr:
Hi, I have an inconvenience installing Nexcloud (over Apache) on a server with IPv6 address.

With my variable `nextcloud_trusted_domain` defined at this way:
```yaml
nextcloud_trusted_domain:
  - "{{ ansible_fqdn }}"
  - "{{ ansible_default_ipv4.address }}"
  - "{{ ansible_default_ipv6.address }}"
```
The Nextcloud site conf file is defined like this:
```xml
<VirtualHost *:80>
  ServerName my-server.org
  Redirect permanent / https://my-server.org/
</VirtualHost>
<VirtualHost *:80>
  ServerName 200.100.20.10
  Redirect permanent / https://200.100.20.10/
</VirtualHost>
<VirtualHost *:80>
  ServerName 200:100:a::10
  Redirect permanent / https://200:100:a::10/
</VirtualHost>
```
### **Being necessary in the permanent redirection, add the brackets around the IPv6:**
```
  Redirect permanent / https://[200:100:a::10]/
```

Something similar happens in `config/config.php`:
```php
  'trusted_domains' => 
  array (
    0 => 'my-server.org',
    1 => '200.100.20.10',
    2 => '200:100:a::10',      // Needs to be modified to:  2 => '[200:100:a::10]',
  ),
```
I'm trying to find an adequate solution for the brackets to be added when appropriate in the templates. If I succeed, I send you the proposal.

aalaesar:
Hello @santiagomr 
I think you are looking for this : https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters_ipaddr.html#wrapping-ipv6-addresses-in-brackets
Test it with:
```yaml
nextcloud_trusted_domain:
  - "{{ ansible_fqdn }}"
  - "{{ ansible_default_ipv4.address }}"
  - "{{ ansible_default_ipv6.address | ipwrap }}"
```
should works and can be easily merged to the templates.

santiagomr:
Good tip. `ipwrap` solves the problems, but can not be used directly in the` nextcloud_trusted_domain` definition. If we do, the apache configuration file is generated like this:
```xml
<VirtualHost *:80>
  ServerName [200:100:a::10]
  Redirect permanent / https://200:100:a::10/
</VirtualHost>
```
The brackets in `ServerName` are a syntax error that doesn't let Apache start

[...]

PR #66, opened 2019-06-04
 Using ipwrap on templates to add brackets when it is needed 

santiagomr:
According to what was discussed in #65, I send you the PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_install:
   - sudo easy_install pip
   # Install ansible
   - sudo pip install -I "ansible${ansible_version}"
-  - sudo pip install netaddr
+  - sudo apt-get install python-netaddr
 
 
 install:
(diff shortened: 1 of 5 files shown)


aalaesar:
Hi @santiagomr 
:+1: 
If you could add ipwrap also to the nginx plugin it would be perfect :smile: 
(I tried to push the commit on the PR but it failed)
nginx_nc.j2:
```yaml
server_name {{ nextcloud_trusted_domain | ipaddr | join(' ') }};  # x2
```

santiagomr:
@aalaesar, I did the rebase on the latest commit of master and added ipwrap to the nginx template.

Finally, I added `netaddr` in the `.travis.yml`, but an error occurs during the build. Maybe you know better why.

aalaesar:
Looked over the error msg. 
Looks like an incompatibility issue between dist python packages and pypi packages and versions gap. (This is Ubuntu 16.04 after all)
It would've probably safer to install via apt python-netaddr

santiagomr:
Ok, I change it from pip to apt

santiagomr:
Yes, switching to apt solved the problem.

PR #75, opened 2019-10-26
Security issue in NGINX/php-fpm

Haelle:
Update nginx conf due to this security issue : https://nextcloud.com/blog/urgent-security-issue-in-nginx-php-fpm/

I'm not an expert of php-fpm so I've just applied the recommended corrections. I did the update on my own configuration and it works.
If something goes wrong on my side I'll inform you.


aalaesar:
Thank you for the PR.  :+1:
Didn't had time to make the patch yet

PR #76, opened 2019-10-28
mitigate CVE-2019-11043

egmont1227:
see https://nextcloud.com/blog/urgent-security-issue-in-nginx-php-fpm/ for futher information.


egmont1227:
duplicate, see #75

PR #77, opened 2019-10-30
Multi os

egmont1227:
this is to support different OS versions.

PHP version is abstracted to variables depending on OS.

It also adds some vagrant jobs to test different OSes. Please feel free to add this to travis, too.


aalaesar:
Thank  you @egmont1227 for sharing this awesome work !

PR #81, opened 2019-11-19
Adds support for Debian 10 Buster

### LINKED ISSUE #80: Support Debian Buster
The role currently supports:

* Ubuntu Trusty (EOL)
* Ubuntu Xenial (LTS)
* Ubuntu Bionic (LTS)
* Debian Stretch

Given that Debian Buster [was released](https://www.debian.org/News/2019/20190706) in July 2019, it's a sound stable target for this role.

conorsch:
Adds support for Debian Buster, via the existing vars import structure. Mostly the change is in the form of PHP7.0 -> PHP7.3. There's a special case on the `php-<version>-mcrypt` package, which does not exist in the Buster repos. On Debian Buster, installation of that package (and only that package) will be skipped.

Closes #80.


aalaesar:
Hi there , thank you for the pull request 
I just made a test install on Buster on my side this week end and noted the difference for php-mcrypt.

The story seems to be: libmcrypt is not maintained since a decade and has been marked as obsolete in php-7.1 and removed since php 7.2. So the package shouldn't even exist in other linux distributions.

[This artice explain quickly why](https://lukasmestan.com/install-mcrypt-extension-in-php7-2/)

So either we should change the condition from php 7.3 to > 7.2 or we move php-mcrypt to the specific packages for php 5 and maybe php 7.0 and 7.1  [but the manual doesn't make reference to this php module since nextcloud 15 at least](https://docs.nextcloud.com/server/15/admin_manual/installation/source_installation.html#prerequisites-for-manual-installation)

conorsch:
> So either we should change the condition from php 7.3 to > 7.2 or we move php-mcrypt to the specific packages for php 5 and maybe php 7.0 and 7.1

Sounds good. If you've no strong preference, then I'll take the latter approach, so folks running on older platforms will still pull in the package.

conorsch:
@aalaesar Pushed up a change that should accommodate. The role had already managed the PHP5 & 7.0 declarations well, so the diff is quite small in the second commit. 

Let me know if you'd like any further changes!

conorsch:
Thanks for your helpful review, @aalaesar, and for maintaining a great role that's saved us a ton of time! :smiley:

PR #85, opened 2019-12-20
*Fix* Debian does not have a "mysql-server" package

privateseabass:
This creates a check for if the user has Debian or not.
In Debian, the package name for mysql is "default-mysql-server".

### DIFF
--- a/tasks/db_mysql.yml
+++ b/tasks/db_mysql.yml
@@ -1,17 +1,9 @@
 ---
-- name: "[mySQL] - Service is installed."
-  package:
-    name: "{{ nextcloud_db_backend }}-server"
-    state: present
-  register: nc_mysql_db_install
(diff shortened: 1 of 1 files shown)


aalaesar:
Hi there,
thank you for the PR !
I would like to keep the number of tasks low
I suggest reusing the initial task by using jinja2's inline condition instead of declarative when: directive.
something like:
```
- name: "[mySQL: Debian] - Service is installed."
  package:
    name: "{{ 'default-' if (ansible_distribution|lower) == "debian" else '' }}{{ nextcloud_db_backend }}-server"
    state: present
  register: nc_mysql_db_install
```
regards

privateseabass:
Kind of like this?

aalaesar:
nicely done 👍

PR #88, opened 2020-03-04
"php_version" to "php_ver" in README

privateseabass:
I changed ever place it said php_version to php_ver to accurately tell the user what variable is used in the playbook.


privateseabass:
Not sure why it failed... I only changed the README.

aalaesar:
> Not sure why it failed... I only changed the README.

Travis CI fail only on Postgresql cases, PG service does not start.
Seems to be be a test regression on my side.
I'll look into it

PR #91, opened 2020-03-29
Moved configuration files to nginx conf folder

mivek:
Depending on the nginx installation it is possible that the folders `sites-available` and `sites-enable` do not exist.
Nginx provides a folder `conf.d` in which it is possible de store configuration files.
I moved the destination from `/etc/nginx/sites-available` to `/etc/nginx/conf.d`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ before_install:
   # Install ansible
   - sudo pip install -I "ansible${ansible_version}"
   - sudo apt-get install python-netaddr
+  - sudo ansible-galaxy install nginxinc.nginx
 
 
 install:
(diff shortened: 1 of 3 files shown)


staticdev:
@aalaesar please merge this one, old nginx config does not work.

aalaesar:
hello @mivek,  @staticdev
Thank you for the contribution,
I' ve checked that debian 10 is still using the legacy `/etc/nginx/sites-available`
so changing the hardcoded is not the proper solution as debian 10 still has 2 years of support ahead.
this need to use some variable in ./vars to reference the proper folder given the context debian/ubuntu  and version.

mivek:
@aalaesar 

I think the location depends on the version of nginx installed. In recent versions configurations files are in `/etc/nginx/conf.d`.
Maybe we could have a variable in `./default` `nextcloud_nginx_configuration` defaulting to `/etc/nginx/sites-available` so we can override it in inventory?

PR #98, opened 2020-04-14
Add indices after nextcloud is installed

mivek:
After nextcloud is installed indices in database could be missing
I added a task to run an `occ` command that adds those missing indices

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -97,3 +97,11 @@
     - nextcloud_apps is defined
     - nextcloud_apps is mapping
   tags: install_apps
+
+- name: Add indices
+  command: php occ db:add-missing-indices
+  args: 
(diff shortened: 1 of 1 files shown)


zeus86:
this check is not idempotent, it always changes, because the output of the occ command is as follows:
```
root@nextcloud1:~# sudo -u www-data php /var/www/nextcloud/occ db:add-missing-indices
Check indices of the share table.
Check indices of the filecache table.
Check indices of the twofactor_providers table.
Check indices of the login_flow_v2 table.
Check indices of the whats_new table.
Check indices of the cards table.
Check indices of the cards_properties table.
Check indices of the calendarobjects_props table.
Check indices of the schedulingobjects table.
Check indices of the oc_properties table.
Done.
```

PR #105, opened 2020-12-01
Allow to use custom "php_ver" e.g. for PHP7.4 on Ubuntu 18.04 with ondrej PPA

klemens-u:
Nextcloud v21 will at least need PHP 7.3. So Ubuntu systems older than 20.04 (which by the way is not supported by this role at the moment) need to upgrade their PHP version. The most common way is to use the ondrej PPA.
https://launchpad.net/~ondrej/+archive/ubuntu/php

While testing this, I noticed that it is not possible to give the Ansible variable "php_ver" from outside of the role. It is overwritten by the task "[ENV] include OS static vars" in setup_env.yml.

The second thing I noticed is that "ondrej PPA" makes quite a mess if you don't give the exact PHP version for a package.  If you use a generic package name like "apt install php-imap" it will install the package for all supported PHP versions (5.6, 7.0, 7.1, 7.2, 7.3, 7.4, 8.0)
I fixed that in php7.4_env.yml. It could (and should) be replicated for other PHP versions.

Both issues are fixed in this pull request.

### How I use it

I use this role to install PHP7.4 on Ubuntu 18.04.
Before the install_nextcloud role, I setup the PPA and uninstall old PHP packages:

```
- name: check php version for upgrade to PHP 7.4
  shell: "php --version | head -n 1 | cut -d ' ' -f 2 | cut -c 1,3"
  changed_when: false
  register: installed_php_version
  # eg.  "72"

- name: uninstall old php packages
  shell: "apt --yes remove 'php*'"
  when: installed_php_version.stdout != "74"

- name: add ondrej ppa for PHP 7.4
  apt_repository:
    repo: ppa:ondrej/php
    state: present
```

After the install_nextcloud role I cleanup a little PHP version mess with php-apcu:

```
# Note: Ubuntu 18.04 with ondrej PPA makes a mess here. Example: php7.4-apcu 
# also installs php-apcu-bc, which in return installs php-apcu-bc for PHP
# 7.0, 7.1, 7.2 and 7.3
# Workaround would be to do a manual "apt install php7.4-apcu php7.4-apcu-bc"

- name: Check for wrong version php-apcu-bc
  shell: "apt list --installed | grep php7.0-apcu-bc"
  failed_when: false
  changed_when: false
  register: installed_apcu
  # grep return code if found = 0

- name: Cleanup php apcu mess
  shell: "apt --yes remove 'php*-apcu*'"
  # from role nextcloud_pre
  when: installed_apcu.rc == 0

- name: Reinstall only php7.4-apcu php7.4-apcu-bc
  shell: "apt --yes install php7.4-apcu php7.4-apcu-bc"
  # from role nextcloud_pre
  when: installed_apcu.rc == 0
```

Furthermore I switch to PHP FPM. It would be very nice if the role could support this in itself, but that's another topic.


aalaesar:
Awesome !  
Thank you for the PR !
good work.

klemens-u:
Thanks. Btw, here are the missing php_envs.yml from the defaults/ directory.
[php_envs.zip](https://github.com/aalaesar/install_nextcloud/files/5671401/php_envs.zip)

PR #113, opened 2021-10-20
Credentials save in MySQL installation

m600x:
- Add credentials saving in mariadb cnf file

When running the playbook the first time, the credentials are created, saved locally in a file and set on the db. But the second time, the task https://github.com/aalaesar/install_nextcloud/blob/master/tasks/db_mysql.yml#L23 will take in account a config file defined in `mysql_credential_file[(ansible_os_family|lower)]` that doesn't not have the generated password (or a provided one via `nextcloud_mysql_root_pwd`) thus failing the playbook.

This task will ensure that the password is set in the same file used to connect to the database.

### DIFF
--- a/tasks/db_mysql.yml
+++ b/tasks/db_mysql.yml
@@ -19,6 +19,15 @@
     nextcloud_mysql_root_pwd: "{{ lookup( 'password', 'nextcloud_instances/'+ nextcloud_instance_name +'/mysql_root.pwd' ) }}"
   when: nextcloud_mysql_root_pwd is not defined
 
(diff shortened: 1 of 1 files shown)


aalaesar:
hello @m600x 
good catch !
What was the distribution with this issue ?
thank you for the PR

m600x:
Hello @aalaesar ,

The issue has been seen under Debian 10.

We have the habit to run multiple time the same playbook on our infrastructure to ensure an exact state free of manual change over time hence why when the second run has been launched the issue raised.

Best regards,

PR #116, opened 2022-02-21
Fix: Remove assets directory

wiktor2200:
I've removed assets directory, because in newer versions it causes a problem with updater.
It was removed from source quite long time ago and it's not used anymore.
Reference:
* https://github.com/nextcloud/server/issues/2610#issuecomment-266275947
* https://github.com/nextcloud/updater/issues/23
* https://help.nextcloud.com/t/the-following-extra-files-have-been-found-assets/21026

### DIFF
--- a/tasks/nc_installation.yml
+++ b/tasks/nc_installation.yml
@@ -162,7 +162,6 @@
     state: directory
   with_items:
     - apps
-    - assets
     - config
     - themes
     - updater


barrelful:
@aalaesar any change of creating a new release on ansible galaxy with this? I have an update problem that should be solved by this change here

nc-update-error

Issue #73, opened 2019-08-27
The PHP memory limit is below the recommended value of 512MB.


mihailstoynov:
The error:
The PHP memory limit is below the recommended value of 512MB.

The solution:
Change
memory_limit = 128M
to
memory_limit = 1G

In `/etc/php/7.2/fpm/php.ini`
then restart:
`systemctl restart php7.2-fpm.service && systemctl restart nginx`

Probably should be a configurable variable in the role?

aalaesar:
Indeed,
I think php settings need refactoring:

- [ ] remove obsolete version (drop support for ubuntu 14, debian 8)
- [ ] simplify php definition on the role to allow easy multi php version support
- [ ] allow any custom php parameter
- [ ] application isolated php settingsmulti version

staticdev:
This issue can be closed on the next release, since current release has a fix but is broken.

PR #122, opened 2022-07-03
Set memory limit on apache / Remove old OS

### LINKED ISSUE #83: fix php memory_limit
>whenever going to Settings > Overview I keep getting the following error: "The PHP memory limit is below the recommended value of 512MB.

Source: https://help.nextcloud.com/t/the-php-memory-limit-is-below-the-recommended-value-of-512mb/42719

To fix it I do something outside the role to fix the php memory_limit:

```
- name: fix php memory limit
  lineinfile:
    path: /etc/php/7.2/fpm/php.ini
    regexp: ^memory_limit = .*$
    line: memory_limit = 512M
  notify: restart php7.2-fpm
```

You can incorporate it in the role.

staticdev:
* Added memory limit to php config for apache (similar to what already has on nginx)
* As mentioned in #73:
   * Removed Ubuntu 14 (trusty) and added 20 instead (focal) also removing old compatibility checks on tasks/php_install.yml
   * Removed Debian 8 (jessie) and added Debian 10 (buster)

Closes #83


tasks/http_apache.yml
```diff
@@ -13,15 +13,16 @@
...
+    backrefs: true
   with_items:
-    - { regexp: 'opcache.enable=0', line: 'opcache.enable=1' }
-    - { regexp: 'opcache.enable_cli', line: 'opcache.enable_cli=1' }
-    - { regexp: 'opcache.interned_strings_buffer', line: 'opcache.interned_strings_buffer=8' }
-    - { regexp: 'opcache.max_accelerated_files', line: 'opcache.max_accelerated_files=10000' }
-    - { regexp: 'opcache.memory_consumption', line: 'opcache.memory_consumption=128' }
-    - { regexp: 'opcache.save_comments', line: 'opcache.save_comments=1' }
-    - { regexp: 'opcache.revalidate_freq', line: 'opcache.revalidate_freq=1' }
+    - {regexp: 'opcache.enable=0', line: 'opcache.enable=1'}
```
wiktor2200, line 18: Is ansible-lint suggested removing trailing spaces after { and before } ?

wiktor2200:
The same applies to Vagrantfile is it used anywhere now? I can see that last travisCI pipeline was triggered 2 years ago.

tasks/http_apache.yml
staticdev: Exactly, and apparently it increases the score in ansible galaxy.

staticdev:
> I can see .gitlab-ci file modified, it hasn't been changed for ages. @aalaesar maybe it's worth deleting it at all? Plus pipeline badge from README (when I clicked it, it shows some Gitlab instance and pipeline which last run 4 years ago).

You are completely right.. I was too focused on references of old Ubuntu and Debian that ended up modifying it. I removed the `.gitlab.-ci` and the badge from README in my latest commit.

staticdev:
> The same applies to Vagrantfile is it used anywhere now? I can see that last travisCI pipeline was triggered 2 years ago.

Travis is still possible to use, but they changed from travis.org to travis.com, maybe some administrative login needs to be done to update it. Another option is to implement a Github Actions pipeline, which may be the most appropriate idea.

wiktor2200:
So let's wait for @aalaesar for an opinion on that. 
I was thinking about Github Actions pipeline in the past for this repo, but have not enough time to contribute.

wiktor2200:
I've added PR for Github Action with ansible-lint, we will see how it's gonna work: https://github.com/aalaesar/install_nextcloud/pull/127/files

staticdev:
@aalaesar latest versions of NC also complain about this. Could you see if we can merge?

staticdev:
> travis

I rebased the PR and also deleted the old Travis pipeline, the Vagrant files can maybe be reused by Molecule.

aalaesar (approved):
LGTM

PR #133, opened 2022-07-30
Remove 10 char limitation for webadmin

staticdev:
10 characters are considered weak, default ansible passwords have 20 characters now (and Nextcloud supports it).


aalaesar (approved):
LGTM

PR #134, opened 2022-08-06
Add Support for Ubuntu 22.04 (Jammy)

maxhoesel:
As the title states, this PR adds support for the newest Ubuntu LTS release. As 22.04 is the first major distro to ship with PHP 8, a few modifications were necessary to accommodate upstream changes. I tested this fork with a local install and everything seems to work just fine (although I haven't tested every feature that this role offers).


aalaesar (approved):
LGTM

PR #136, opened 2022-08-06
Add initial molecule tests

staticdev:
- Create default molecule case with default values from the role.
- Add metadata for ansible-galaxy.
- Set name as `aalaesar.nextcloud` since install is the default action for ansible-roles.

Tests are added by they don't pass since there is a bug https://github.com/aalaesar/install_nextcloud/issues/137


staticdev:
@wiktor2200 can you check this PR? I don't use mysql, but seems that if you install this role using all default values we get this error.

wiktor2200:
Hi! I've taken a look on that and the problem here seems to be lack of `nextcloud_mysql_root_pwd` variable set.
It's commented out in defaults (since early beginning - init commit) and I suppose that was done for security reasons.
https://github.com/aalaesar/install_nextcloud/blob/master/defaults/main.yml#L90
And second - generating it with this step also seems not to work:
```
   TASK [aalaesar.nextcloud : [mySQL] - generate mysql root Password:] ************
  ok: [instance]
```

So I guess the best option here would be overwriting this variable in molecule definition or fixing this generator. 

BTW this generator is used 4 times in whole repo so I guess the same problem would be for postgres and web_admin:
* https://github.com/aalaesar/install_nextcloud/blob/b8ce19b4dc42679be0a4161ead41db882dde2765/tasks/db_mysql.yml#L19
* https://github.com/aalaesar/install_nextcloud/blob/b8ce19b4dc42679be0a4161ead41db882dde2765/tasks/db_mysql.yml#L19
* https://github.com/aalaesar/install_nextcloud/blob/3f4527ffbf3486e89033e4e58c8820262387282a/tasks/db_postgresql.yml#L14
* https://github.com/aalaesar/install_nextcloud/blob/d91c69d8b18eed594f16460787834c49665a8838/tasks/nc_installation.yml#L18-L19

I've created PR for fixing generator in your repo in branch from which this PR was done: https://github.com/staticdev/install_nextcloud/pull/1

staticdev:
> Hi! I've taken a look on that and the problem here seems to be lack of `nextcloud_mysql_root_pwd` variable set. It's commented out in defaults (since early beginning - init commit) and I suppose that was done for security reasons. https://github.com/aalaesar/install_nextcloud/blob/master/defaults/main.yml#L90 And second - generating it with this step also seems not to work:
> 
> ```
>    TASK [aalaesar.nextcloud : [mySQL] - generate mysql root Password:] ************
>   ok: [instance]
> ```
> 
> So I guess the best option here would be overwriting this variable in molecule definition or fixing this generator.
> 
> BTW this generator is used 4 times in whole repo so I guess the same problem would be for postgres and web_admin:
> 
>     * https://github.com/aalaesar/install_nextcloud/blob/b8ce19b4dc42679be0a4161ead41db882dde2765/tasks/db_mysql.yml#L19
> 
>     * https://github.com/aalaesar/install_nextcloud/blob/b8ce19b4dc42679be0a4161ead41db882dde2765/tasks/db_mysql.yml#L19
> 
>     * https://github.com/aalaesar/install_nextcloud/blob/3f4527ffbf3486e89033e4e58c8820262387282a/tasks/db_postgresql.yml#L14
> 
>     * https://github.com/aalaesar/install_nextcloud/blob/d91c69d8b18eed594f16460787834c49665a8838/tasks/nc_installation.yml#L18-L19
> 
> 
> I've created PR for fixing generator in your repo in branch from which this PR was done: [staticdev#1](https://github.com/staticdev/install_nextcloud/pull/1)

@wiktor2200 I have a different suspect, it looks like the problem is just with the root when it generates. I tried using the other generators and overwriting the password for the first one and the installation when through until the end.

wiktor2200:
Maybe sth was changed since first release? I can see in Actions that mysql has problems with assigning address and getting `root @ localhost`, so maybe there is a need to add: `become: true` there?

staticdev:
@aalaesar @wiktor2200 it is safe to merge this PR, it is only the molecule tests. It is detecting some bugs with the default configuration. I already fixed the mysql related (here), but now there is a new one related to cron.

aalaesar:
hello @staticdev 
thank you for putting those effort into the CI 👍 
> * Set name as `aalaesar.nextcloud` since install is the default action for ansible-roles.
I have already taken the name for the ansible collection I'm slowly working on. so i'll decline the name change for now

I'll push the beginning of the collection soon in a new repository

staticdev:
@aalaesar understood, no problem. I have just reverted it.

README.md
```diff
@@ -1,7 +1,7 @@
 [![Ansible-lint status](https://github.com/aalaesar/install_nextcloud/actions/workflows/ansible-lint.yml/badge.svg)](https://github.com/aalaesar/install_nextcloud/actions?workflow=Ansible%20Lint)
 [![YAML-lint status](https://github.com/aalaesar/install_nextcloud/actions/workflows/yamllint.yml/badge.svg)](https://github.com/aalaesar/install_nextcloud/actions?workflow=Yaml%20Lint)
 
-# install_nextcloud
+# Ansible role: Nextcloud
```
aalaesar, line 4: Since keeping the old name for now, can you also revert that or set the old name as 
```suggestion
# Ansible role: install_nextcloud
```
?

Thank you :)

aalaesar (changes requested):

aalaesar (approved):

Issue #112, opened 2021-05-10
Task "Add indices" is always changing


apatard:
While testing this role, it turns out that the task "Add indices" is always reporting a change, even if nothing is changed.

Looks like the 'db:add-missing-indices' command does something only if needed and from the source code, it's writing 'Done' to stdout only if nothing has been done : https://github.com/nextcloud/server/blob/ca49ba644141e1316d98c8476510d0b8ffceb1b3/core/Command/Db/AddMissingIndices.php#L315
So, a quick fix would be something like:
```
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -102,6 +102,8 @@
   command: php occ db:add-missing-indices
   args: 
     chdir: "{{ nextcloud_webroot }}"
+  register: nc_indices_cmd
   become_user: "{{ nextcloud_websrv_user }}"
   become: yes
   become_flags: "{{ ansible_become_flags | default(omit) }}"
+  changed_when: '"Done" not in nc_indices_cmd.stdout'
```
Not sure if it's working when the indices are missing.

PR #138, opened 2022-08-07
Fix bug on mysql passwords

### LINKED ISSUE #137: Mysql error on installation using defaults
Error discovered using molecule tests.

```
   TASK [aalaesar.nextcloud : [mySQL] - Update mysql root password] ***************
  fatal: [instance]: FAILED! => {"changed": false, "msg": "unable to connect to database, check login_user and login_password are correct or /etc/mysql/debian.cnf has the credentials. Exception message: (2003, \"Can't connect to MySQL server on 'localhost' ([Errno 99] Cannot assign requested address)\")"}
```

staticdev:
Depends on molecule PR https://github.com/aalaesar/install_nextcloud/pull/136

Closes #137

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,12 +69,12 @@ nextcloud_get_latest: true
 Specify if the "latest" archive should be downloaded.
 
 ```YAML
-# nextcloud_version_major: 10
+# nextcloud_version_major: 24
 ```
 Specify what major version you desire.
 
 ```YAML
-# nextcloud_version_full: "10.0.3"
+# nextcloud_version_full: "24.0.3"
 ```
 The full version of the desired nextcloud instance. type **M.F.P** _(Major.Feature.Patch)_
 
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -1,6 +1,7 @@
 ---
 - name: Converge
   hosts: all
+  become: true
 
   pre_tasks:
     - name: Update apt cache.
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -5,7 +5,7 @@
 
   pre_tasks:
     - name: Update apt cache.
-      apt: update_cache=yes cache_valid_time=600
+      ansible.builtin.apt: update_cache=yes cache_valid_time=600
       when: ansible_os_family == 'Debian'
       changed_when: false
 
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -5,7 +5,9 @@
 
   pre_tasks:
     - name: Update apt cache.
-      ansible.builtin.apt: update_cache=yes cache_valid_time=600
+      ansible.builtin.apt:
+        update_cache: true
+        cache_valid_time: 600
       when: ansible_os_family == 'Debian'
       changed_when: false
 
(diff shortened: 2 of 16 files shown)


tasks/setup_env.yml
```diff
@@ -57,3 +57,8 @@
   when:
     - nc_sudo_installed_result.rc is defined
     - nc_sudo_installed_result.rc != 0
+
+- name: "[ENV] - Define mysql_daemon."
+  set_fact:
```
wiktor2200, line 62:
```suggestion
  ansible.builtin.set_fact:
```

tasks/db_mysql.yml
```diff
@@ -12,7 +12,11 @@
   vars:
     nc_mysql_deps:
       - "php{{ php_ver }}-mysql"
-      - "python{{ '3' if ansible_python.version.major == 3 else '' }}-pymysql"
+      - "python3-pymysql"
+
+- name: Ensure MySQL is started and enabled on boot.
+  service: "name={{ mysql_daemon }} state=started enabled={{ nextcloud_db_enabled_on_startup }}"
```
wiktor2200, line 18: You should move this definition to handlers and use this notation:
```suggestion
  service: 
    name={{ mysql_daemon }}
    state=started
    enabled={{ nextcloud_db_enabled_on_startup }}"
```

molecule/default/converge.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: Converge
+  hosts: all
+  become: true
+
+  pre_tasks:
+    - name: Update apt cache.
+      ansible.builtin.apt: update_cache=yes cache_valid_time=600
```
wiktor2200, line 8:
```suggestion
      ansible.builtin.apt: 
        update_cache: true
        cache_valid_time: 600
```

wiktor2200 (changes requested):
Please change to Ansible YAML syntax.

aalaesar:
Hello @staticdev.
I see some issues  with the mysql authentication on the system.
Since the 6 years I've made this role I found that passing the password as environment  variable is secured enough : it doesn't not to appear in ps command and in ansible very verbose logs so I dumped the .my.cnf configuration on my others roles.
Maybe this would resolve the mess as the role knows the root password during the mysql installation.
How about I implement that ? This should simplify this part.

staticdev:
> Hello @staticdev. I see some issues with the mysql authentication on the system. Since the 6 years I've made this role I found that passing the password as environment variable is secured enough : it does not to appear in ps command and in ansible very verbose logs so I dumped the .my.cnf configuration on my others roles. Maybe this would resolve the mess as the role knows the root password during the mysql installation. How about I implement that ? This should simplify this part.

Don't bother I am on it.. I refactored the password generation to be done on `setup_env`, this has some advantages and enables us to use roles that need password as a parameter.

staticdev:
@aalaesar Mysql is also fixed.. the missing error is the one reported on https://github.com/aalaesar/install_nextcloud/issues/112
I can try to fix in a later PR.

README.md
```diff
@@ -32,7 +32,7 @@ This role requires root access, so either configure it in your inventory files,
 - hosts: dnsserver
   become: true
   roles:
-    - role: aalaesar.install_nextcloud
+    - role: aalaesar.nextcloud
```
aalaesar, line 35: seems the name change was also commited in this branch.
can you revert it ? thanks :)

aalaesar (changes requested):

README.md
staticdev: @aalaesar this PR was synced automatically after the merge please check again.

PR #141, opened 2022-08-14
Molecule improvements

### LINKED ISSUE #112: Task "Add indices" is always changing

staticdev:
- Add idempotent install task
- Fix buster daemon timeout
- Add molecule badge to README
- Molecule uses version 23 since 24 requires php 7.4 (and Debian 10 has 7.3)

Closes #112 
Superseeds #140

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,6 +46,15 @@ jobs:
         run: |
           python3 -m pip install --constraint=.github/workflows/constraints.txt ansible molecule[docker] docker netaddr
 
+      # See: https://github.com/geerlingguy/ansible-role-mysql/issues/422
+      - name: Disable AppArmor on Debian.
+        run: |
+            set -x
+            sudo apt-get install apparmor-profiles
+            sudo ln -s /etc/apparmor.d/usr.sbin.mysqld /etc/apparmor.d/disable/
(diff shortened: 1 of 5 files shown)


staticdev:
@aalaesar I took the liberty to get your change to idempotency in this PR. Finally we have molecule working for all OS with idempotency =)

PR #142, opened 2022-08-15
add reference configurations for OS and PHP contextual variables

aalaesar:
## reduce maintenance complexity by adding reference configurations:
removes the need to use multiple vars-files per version of PHP and debian/ubuntu. 
configuration is centralised in 2 yaml files in defaults/
## reduced configuration:
OS and PHP vars are defined with defaults value for common vars.
So that it is only required to add what's changing from defaults for new  OS and php versions.
## reduced task number
everything is loaded with lookups and filters instead of include_vars tasks
## completely retro-compatible.
- only 2 additional vars not exposed to the users, all other vars are keeping their original name.
- all vars using the new references are defined in defaults/main.yml so users can still change their php installation, as documented in readme


aalaesar:
@staticdev
Hi. I'm a bit confused with the daemon_name variable.
why is it required per distribution ? I 've developed the question in #143.
Can you give me some context ?
thanks 😃

staticdev:
> @staticdev Hi. I'm a bit confused with the daemon_name variable. why is it required per distribution ? I 've developed the question in #143. Can you give me some context ? thanks smiley

I am answering there then.

defaults/php_config_ref.yml
```diff
@@ -0,0 +1,48 @@
+---
+# define parameters depending for the php version installed
+# for any new version, only specify keys with values different from defaults
+# defaults should work for php > 7.3
+defaults:
+  php_dir: "/etc/php/{{ php_ver }}"
+  php_bin: "php-fpm{{ php_ver }}"
+  php_pkg_apcu: "php{{ php_ver }}-apcu"
```
staticdev, line 8: You can replace php_pkg_apcu by:

if version < 7.3 it is php-apcu otherwise it is php{{ php_ver }}-apcu.

defaults/php_config_ref.yml
```diff
@@ -0,0 +1,48 @@
...
+    - "php{{ php_ver }}-bcmath"
+    - "php{{ php_ver }}-gmp"
+    - "php{{ php_ver }}-imagick"
+    - "php{{ php_ver }}-mbstring"
+    - "php{{ php_ver }}-redis"
+    - "php{{ php_ver }}-xml"
+    - "php{{ php_ver }}-zip"
+  php_socket: "/run/php/php{{ php_ver }}-fpm.sock"
+
+7_3:
```
staticdev, line 19: I would even remove everything from this line on and replace with 2 tasks:

if php_ver <= 7_3 you need to install:
    - "php-imap"
    - "php-imagick"
    - "php-redis"
    - "php{{ php_ver }}-xml"
    - "php{{ php_ver }}-zip"
    - "php{{ php_ver }}-mbstring"

If it is 7.0 you add a new task just to add "php7.0-mcrypt" -> this task should be removed soon as Nextcloud is phasing out support vor versions with php 7.0.

staticdev (changes requested):

defaults/php_config_ref.yml
aalaesar: Thank you for feedback @staticdev 👍 
I understand your point: we could go even further in automation by templating even more the reference configuration for PHP.
I won't add tasks for managing special cases. The goal of the pr is to remove them and reduce   the tasks to only "work" things.
I think templating the supported php conf  will make the role harder to apprehend for newcomers.
Simply stating ( not that much ) config keys that change from default will look simpler than a bunch of if blocks in the defaults key.
It's a compromise between  ease of maintenance  and features I think.

I'll make a few more change to remove more tasks in setup_env in this PR
Do not hesitate to counter me on that.
Regards.
Aal

staticdev: @aalaesar this is ok, then I would say we try to merge this PR before introducing new changes. It already has a good cleanup size. In this case I would recommend adding new changes in a new PR if it is not too annoying for you.

defaults/php_config_ref.yml
aalaesar: Actually it  also works for php7.4

staticdev (approved):

PR #148, opened 2022-10-17
Move the installation of the PHP database clients to php_install

jspaans91:
Moved the installation of the PHP Database clients to php_install.yml instead of the db installation, before when nextcloud_install_db is set to false the PHP Database clients are not installed.

This is common when you use an external database server.


wiktor2200 (approved):
LGTM

staticdev (approved):

PR #149, opened 2022-10-17
Set crontab 10 minutes instead of 15, because of warning message

jspaans91:
Changed the cron schedule to 10 minutes instead of 15 minutes, when the crontab has not run for more then 10 minutes it creates a warning in Nextcloud (see the screenshot below).

image

### DIFF
--- a/tasks/nc_installation.yml
+++ b/tasks/nc_installation.yml
@@ -132,7 +132,7 @@
     - name: "[NC] - Install Cronjob"
       ansible.builtin.cron:
         name: "Nextcloud Cronjob"
-        minute: "*/15"
+        minute: "*/10"
         user: "{{ nextcloud_websrv_user }}"
         job: "php {{ nextcloud_webroot }}/cron.php"
         cron_file: "nextcloud"


wiktor2200 (approved):
LGTM, It makes sense now.
Back in 2019 there were discussions about huge instances of NC not getting fully compatible with short cron periods.
But now I guess it's solved.
Warning is triggered by this check:
https://github.com/nextcloud/server/blob/master/apps/settings/src/components/BasicSettings/BackgroundJob.vue#L28
and then whole logic is here:
https://github.com/nextcloud/server/blob/master/apps/settings/src/components/BasicSettings/BackgroundJob.vue#L134
and it's checking 600s=10mins, so it makes sense to change it.

staticdev (approved):

staticdev:
Dank je wel @jspaans91 merged!

barrelful:
@aalaesar can we have a new release with this?

staticdev:
@barrelful I expect we will have a release as soon as we have this https://github.com/aalaesar/install_nextcloud/pull/163 ;)

PR #150, opened 2022-10-17
Added task to disable default installed apps

jspaans91:
Added a task to disable apps that are enabled by default and you maybe not want. The variable for the list with apps you want to disable is "nextcloud_disable_apps"

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -105,6 +105,7 @@ nextcloud_tls_session_cache_size: 50m  # in Byte or human readable size notation
 
 # [APPS]
 nextcloud_apps: {}
+nextcloud_disable_apps: {}
 
 # [SYSTEM]
 # nextcloud_mysql_root_pwd: "secret"
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -106,6 +106,33 @@
     - nextcloud_apps is mapping
   tags: install_apps
 
+- name: Disable Nextcloud Apps
+  block:
+    - name: "[NC apps] - lists the number of apps available in the instance."
+      ansible.builtin.command: php occ app:list --output=json_pretty --no-warnings
+      args:
+        chdir: "{{ nextcloud_webroot }}"
+      become_user: "{{ nextcloud_websrv_user }}"
+      become: true
+      become_flags: "{{ ansible_become_flags | default(omit) }}"
+      changed_when: false
+      check_mode: false
+      register: nc_apps_list
+
+    - name: "[NC apps] - convert list to yaml."
+      ansible.builtin.set_fact:
+        nc_available_apps: "{{ nc_apps_list.stdout | from_json }}"
+
+    - name: "[NC apps] - Disable apps"
+      include_tasks: ./nc_disable_apps.yml
+      # do if the app is  enabled and ( (archive path is not "") or (app is disabled) )
+      when:
+        - item in nc_available_apps.enabled
+      with_items: "{{ nextcloud_disable_apps }}"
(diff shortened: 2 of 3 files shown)


jspaans91:
@aalaesar sorry for bothering you, but can you check if we want this functionality be added to this role or not?

If not, I'll add it to my own Nexcloud customization role.

Cc @staticdev

aalaesar:
> @aalaesar sorry for bothering you, but can you check if we want this functionality be added to this role or not?
> 
> If not, I'll add it to my own Nexcloud customization role.
> 
> Cc @staticdev

hello @jspaans91  I was reviewing the code at the moment

> @aalaesar sorry for bothering you, but can you check if we want this functionality be added to this role or not?
> 
> If not, I'll add it to my own Nexcloud customization role.
> 
> Cc @staticdev

hello @jspaans91  and thank you for the idea and the contribution
I was reviewing the PR ealier and I think your idea is great, as nextcloud is coming with more and more shipped application. One may not want/need them all.

As the role is a little big, I tend to look for minimising the number of tasks to do things. 🤔 
The two main tasks to do it are here, there is also some refinement required to simplify how to use the list of apps returned by the occ cli without using tasks.
I can make suggestions if you're ok 😃

tasks/main.yml
```diff
@@ -106,6 +106,33 @@
     - nextcloud_apps is mapping
   tags: install_apps
 
+- name: Disable Nextcloud Apps
```
staticdev, line 109: I think it is not needed to add this session to `main.yml`. A more compact and easy to maintain solution would be to add to the end of `nc_installation.yml` a block that executes `when: nextcloud_disable_apps is defined` to run `php occ app:disable "{{ nc_app_name }}"` with items from this list.

defaults/main.yml
```diff
@@ -105,6 +105,7 @@ nextcloud_tls_session_cache_size: 50m  # in Byte or human readable size notation
 
 # [APPS]
 nextcloud_apps: {}
+nextcloud_disable_apps: {}
```
staticdev, line 108: When adding a new parameter, especially when default is empty, is very good if we have an actual example with a list in the [Example Playbooks](https://github.com/aalaesar/install_nextcloud#example-playbook).

tasks/nc_disable_apps.yml
```diff
@@ -0,0 +1,16 @@
+---
```
staticdev, line 1: With the comment from `main.yml` this makes also this file not necessary.

staticdev (changes requested):

staticdev:
Thanks a lot for the contribution again @jspaans91! I also like the feature. I made some suggestions already, not sure if would be in line with @aalaesar was also thinking. Feel free to comment or propose alternatives.

tasks/main.yml
aalaesar: @staticdev suggestion is even simpler than what I had i mind, so I :+1:  it

1tasks looping on the list and changed status depending of the stdout:
- ok : "No such app enabled: {{ item }}"
- changed: "{{ item }} <a version> enabled"

rc is always 0 in both case

should do something like that:
```yaml
changed_when:
  - "'enabled' in stdout"
  - "'No such app" not in stdout
```

staticdev:
Hoi @jspaans91 just a vrindelijke ping. Did you had opportunity to take a look at our code review? What do you think about it?

staticdev:
@jspaans91 I will closed this PR for inactivity, if you are able to reply us I will be happy to reopen.

PR #153, opened 2022-11-22
Add dependabot

staticdev:
* Dependabot will automatically check dependencies and propose PRs.


staticdev:
Lgtm @aalaesar ?

aalaesar:
@staticdev isn't that tool overkill for a somple ansible role? What is the use case ?

staticdev:
@aalaesar do not think it is overkill. I have it in ALL my Github projects. In this repo, it would find updates in Github Actions which are very frequent and also in versions of libraries we use. I think it is very important to keep the role fresh.

aalaesar:
Ok I understand the usefullness of the tool, thank you for putting light on it 😃 
However, seeing [this](https://github.com/dependabot/dependabot-core/issues/3522), I doubt it is usefull in this case, unless there is a dependency manifest I forgot somewhere.
Or maybe I just can't project the example you gave me because I don't know the tool. 😕  Is it for molecule ?

staticdev:
@aalaesar you are right that it could be better if it had direct support to Galaxy, though it is still useful for dependencies used to run molecule and also to update the existing actions in the workflows. As you see for example in the last commit Labeler action is failing on version 4.0.0, current version is 4.1.0.. it is possible that we will not have problems with the new version. Last but not least it also updates ansible-lint and yamllint that has been updated frequently lately.

aalaesar (approved):
Ok then.

PR #156, opened 2022-11-26
Remove unsuported distros from before 2017

staticdev:
Ubuntu Xenial is from 16.04 and Debian Stretch is also from 2017, both no longer supported.

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -11,16 +11,14 @@ galaxy_info:
   platforms:
     - name: Ubuntu
       versions:
-        - xenial
-        - bionic
-        - focal
         - jammy
+        - focal
+        - bionic
 
     - name: Debian
       versions:
-        - stretch
-        - buster
         - bullseye
+        - buster
 
   galaxy_tags:
     - nextcloud


wiktor2200 (commented):
Old distros are not supported anymore but there are some exceptions.
Both Ubuntu and Debian has "Extended LTS" for commercial enterprise usage and in this case Ubuntu 16.04 (xenial) is supported up to Q1 2026 (https://ubuntu.com/about/release-cycle) and Debian 9 (stretch) to Q2 2027 (https://wiki.debian.org/LTS/Extended).
If we are deciding to drop support for xenial and stretch, it LGTM though.

staticdev:
@wiktor2200 I believe this role already does not work in these versions. They have very old php versions and cannot run newer versions of NC.

aalaesar:
I'm ok with keeping only the distros the CI is running on. Having an enterprise using the role seems really ultra corner case.

aalaesar (approved):
LGTM

staticdev:
In addition to that, we have even Debian 10 running on NC 23 which is almost EOL according to last NC release notes. But this is a topic for a future removal.

PR #162, opened 2022-11-27
Fix labeler in CI - help wanted label warning

wiktor2200:
I've noticed that one of labels has broken definition which was causing "Labeler" fail.
https://github.com/aalaesar/install_nextcloud/actions/runs/3559378521
I've fixed `string` issue and added `"` for all color to avoid such errors in the future.


staticdev (approved):

aalaesar:
Ho! I was trying to understand why this label was failing. Now I get it! Thank you !

PR #163, opened 2022-11-27
Automate release Ansible Galaxy

staticdev:
This is a handy workflow that releases on Ansible Galaxy everytime the version is changed on `pyproject.toml` it also creates a release on Github automatically with the release notes from release-draft.

To make it work it is necessary to create a token on Ansible Galaxy and add it as `GALAXY_TOKEN` in https://github.com/aalaesar/install_nextcloud/settings/secrets/actions/new


aalaesar:
Looks nice.
review ongoing.
I'm trying to understand the code & workflow first :)

staticdev:
Just a missing info, in galaxy.ansible.com you just need to login, go to preferences on the corner and Show API Key to have the secret for Github I mentioned in the description.

key

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+          poetry --version
+
+      - name: Check if there is a parent commit
+        id: check-parent-commit
+        run: |
+          echo "::set-output name=sha::$(git rev-parse --verify --quiet HEAD^)"
+
+      - name: Detect and tag new version
+        id: check-version
+        if: steps.check-parent-commit.outputs.sha
```
aalaesar, line 41: I don't understand where's the condition here 😕

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: salsify/action-detect-and-tag-new-version@v2.0.2
+        with:
+          tag-template: "{VERSION}"
+          version-command: |
+            bash -o pipefail -c "poetry version | awk '{ print \$2 }'"
```
aalaesar, line 46: hmm. 🤔 
why introducing poetry just to use the version feature ?
A simple version text file could also be enough.

staticdev:
@aalaesar I was able to simplify it and remove poetry. We will have to test to see how it goes.

staticdev:
@aalaesar I just tested [in another role of mine](https://github.com/staticdev/ansible-role-signal/pull/133).. it works fine also.

aalaesar:
OK I've added the secret 

Still, I don't understand how the condition works [here](https://github.com/staticdev/install_nextcloud/blob/ci/automate-release-ansible-galaxy/.github/workflows/release.yml#L36)

staticdev:
> OK I've added the secret
> 
> Still, I don't understand how the condition works [here](https://github.com/staticdev/install_nextcloud/blob/ci/automate-release-ansible-galaxy/.github/workflows/release.yml#L36)

@aalaesar it is just a check to see if there is parent commit that has a `current-version.txt`, if it does not have it should not apply version tag.

aalaesar (approved):
LGTM

Issue #165, opened 2022-12-03
Galaxy does not handle correctly branch renaming


staticdev:
Check https://github.com/ansible/galaxy/issues/2535

staticdev:
Looks like it worked.

wiktor2200:
I can see it works, but it's getting only version without `v` preffix on galaxy. I guess it should work anyways, but in same cases may break some dependencies :thinking: 
https://galaxy.ansible.com/aalaesar/install_nextcloud

staticdev:
> I can see it works, but it's getting only version without `v` preffix on galaxy. I guess it should work anyways, but in same cases may break some dependencies, so it's just good to have it in mind thinking https://galaxy.ansible.com/aalaesar/install_nextcloud
> 
> BTW I'd prefer as it is right now without `v` - just version (most of the roles are versioned this way) ;)

I also noticed that, and was going to comment that vast majority of roles in Ansible Galaxy does not a `v`. I consider it an unintentional bugfix XD

PR #166, opened 2022-12-03
Add branch to ansible-galaxy role import

staticdev:
Tentative to solve https://github.com/aalaesar/install_nextcloud/issues/165

@aalaesar I see many people in [this thread](https://github.com/ansible/galaxy/issues/2535) had to recreate master to fix and upload manually, then delete master. Hopefully this will avoid this but I cannot be sure since all my roles already have main branch it is hard to test.


wiktor2200 (approved):
LGTM

PR #170, opened 2022-12-13
Increase php memory_limit to 768M

wiktor2200:
I've just installed fresh Nextcloud 25.0.2 and with default role's values it's getting error: `The PHP memory limit is below the recommended value of 512MB.`
Current default was 512M, so I've changed it to 768M (added 256M), which solved this issue.
@aalaesar @staticdev What do you think about changing default to new value?

### DIFF
--- a/README.md
+++ b/README.md
@@ -440,7 +440,7 @@ Set the size of the shared nginx TLS session cache to 50 MB.
 Install and use a custom version for PHP instead of the default one:
 
 ```yaml
-php_ver: '7.1'
+php_ver: '8.1'
 php_dir: "/etc/php/{{ php_ver }}"
 php_bin: "php-fpm{{ php_ver }}"
 php_pkg_apcu: "php-apcu"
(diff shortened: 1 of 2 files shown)


staticdev:
> I've just installed fresh Nextcloud 25.0.2 and with default role's values it's getting error: `The PHP memory limit is below the recommended value of 512MB.` Current default was 512M, so I've changed it to 768M (added 256M), which solved this issue. @aalaesar @staticdev What do you think about changing default to new value?

I see it as a possible bug on nextcloud actually. Maybe we can double check before changing the role.

wiktor2200:
Hi! Sorry for late response. I've tested it twice again on new, clean VMs and now I cannot reproduce it anymore and I'm not getting any error.

> I think you used ansible as intended by changing the default value.

Yes, I did, copied whole file from defaults and tweak it as always did ;)
I meant that default should work out-of-the-box (maybe not fully optimized, but without errors.) Anyway it vanished, so current defaults are good and error/warnings free on new instances. So I'm closing this PR for now :)

PR #176, opened 2022-12-28
Remove old <7.4 php versions

### LINKED ISSUE #175: Cleanup references to php versions before 7.4
In addition to problems from #174, there are references to even older versions on https://github.com/aalaesar/install_nextcloud/blob/main/defaults/php_config_ref.yml. Let's first remove them.

staticdev:
Closes #175

### DIFF
--- a/defaults/php_config_ref.yml
+++ b/defaults/php_config_ref.yml
@@ -1,7 +1,7 @@
 ---
 # define parameters depending for the php version installed
 # for any new version, only specify keys with values different from defaults
-# defaults should work for php > 7.3
+# should work for php >= 7.4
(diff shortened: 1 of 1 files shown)


wiktor2200 (approved):
LGTM, we will also get rid of warning in ansible-lint for wrong formatting (7_3, 7_2, 7_0).

wiktor2200:
BTW, if we drop support for php < 7.4 and we've already dropped support for NC < 24 maybe it's worth removing molecule test for Debian 10 with NC23, WDYT?

aalaesar (approved):
LGTM

PR #188, opened 2023-01-06
Allow to disable preinstalled apps after installation

staticdev:
Follow-up from https://github.com/aalaesar/install_nextcloud/pull/150
- Add example on README
- Add example on coverge for molecule test

### DIFF
--- a/molecule/default/latest-converge.yml
+++ b/molecule/default/latest-converge.yml
@@ -13,7 +13,7 @@
   vars:
     nextcloud_disable_apps:
       - photos
-      - theming
+      - external
       - mail
   roles:
     - role: aalaesar.install_nextcloud
--- a/molecule/default/latest-converge.yml
+++ b/molecule/default/latest-converge.yml
@@ -13,5 +13,7 @@
   vars:
     nextcloud_disable_apps:
       - photos
+      - theming
+      - mail
   roles:
     - role: aalaesar.install_nextcloud
--- a/molecule/default/latest-converge.yml
+++ b/molecule/default/latest-converge.yml
@@ -10,6 +10,8 @@
         cache_valid_time: 600
       when: ansible_os_family == 'Debian'
       changed_when: false
-
+  vars:
+    nextcloud_disable_apps:
+      - photos
   roles:
     - role: aalaesar.install_nextcloud
--- a/tasks/nc_installation.yml
+++ b/tasks/nc_installation.yml
@@ -176,32 +176,32 @@
     - themes
     - updater
 
-name: "[NC] - Disable Nextcloud apps"
-when: nextcloud_disable_apps is defined
-block:
-  - name: "[NC apps] - Lists the number of apps available in the instance."
-    ansible.builtin.command: php occ app:list --output=json_pretty --no-warnings
-    args:
-      chdir: "{{ nextcloud_webroot }}"
-    become_user: "{{ nextcloud_websrv_user }}"
-    become: true
-    become_flags: "{{ ansible_become_flags | default(omit) }}"
-    changed_when: false
-    check_mode: false
-    register: nc_apps_list
+- name: "[NC] - Disable Nextcloud apps"
+  when: nextcloud_disable_apps is defined
+  block:
+    - name: "[NC apps] - Lists the number of apps available in the instance."
+      ansible.builtin.command: php occ app:list --output=json_pretty --no-warnings
+      args:
+        chdir: "{{ nextcloud_webroot }}"
+      become_user: "{{ nextcloud_websrv_user }}"
+      become: true
+      become_flags: "{{ ansible_become_flags | default(omit) }}"
+      changed_when: false
+      check_mode: false
+      register: nc_apps_list
 
-  - name: "[NC apps] - Convert list to yaml."
-    ansible.builtin.set_fact:
-      nc_available_apps: "{{ nc_apps_list.stdout | from_json }}"
+    - name: "[NC apps] - Convert list to yaml."
+      ansible.builtin.set_fact:
+        nc_available_apps: "{{ nc_apps_list.stdout | from_json }}"
 
-  - name: "[NC apps] - Disable apps"
-    ansible.builtin.command: php occ app:disable "{{ item }}"
-    become_user: "{{ nextcloud_websrv_user }}"
-    become_flags: "{{ ansible_become_flags | default(omit) }}"
-    become: true
-    when:
-      - item in nc_available_apps.enabled
-    changed_when:
-      - "'enabled' in stdout"
-      - "'No such app' not in stdout"
-    with_items: "{{ nextcloud_disable_apps }}"
+    - name: "[NC apps] - Disable apps"
+      ansible.builtin.command: php occ app:disable "{{ item }}"
+      become_user: "{{ nextcloud_websrv_user }}"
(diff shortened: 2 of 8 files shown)


staticdev:
@aalaesar @wiktor2200 I tried to resurrect this feature, but not sure how is the intended use do disable an app. Where any of you successful testing it? Maybe can you help me adding an example of useful/common app to disable?

aalaesar (changes requested):
As I said in [this comment](https://github.com/aalaesar/install_nextcloud/pull/150#discussion_r1013520136) (but I switched disabled with enabled in my example 😨 )
This can be achieved in one task with a loop and I would prefer this simplier implementation:
-  I don't think it is required to fetch the list of app prior to disabling any
-  it is better of readability and idempotence display to loop over all apps declared as "I want this disabled" in the var by having the changed state defined in function of the `occ app:disable` command's stdout/err.

Regards,
Aal

PS: I'll seriously consider doing a module for managing nextcloud apps once the repo is in the nextcloud organization.

tasks/nc_installation.yml
```diff
@@ -175,3 +175,33 @@
...
+      ansible.builtin.command: php occ app:disable "{{ item }}"
+      become_user: "{{ nextcloud_websrv_user }}"
+      become_flags: "{{ ansible_become_flags | default(omit) }}"
+      become: true
+      when:
+        - item in nc_available_apps.enabled
+      changed_when:
+        - "'enabled' in stdout"
+        - "'No such app' not in stdout"
+      with_items: "{{ nextcloud_disable_apps }}"
```
aalaesar, line 207: As described in my previous comments, this only task _should_ be enough to do the disable step
```suggestion
    - name: "[NC apps] - Disable apps"
      ansible.builtin.command: php occ app:disable "{{ item }}"
      become_user: "{{ nextcloud_websrv_user }}"
      become_flags: "{{ ansible_become_flags | default(omit) }}"
      become: true
      register: _occ_app_disable_result
      changed_when:
        - "'disabled' in _occ_app_disable_result.stdout"
        - "'No such app' not in _occ_app_disable_result.stdout"
      with_items: "{{ nextcloud_disable_apps }}"
```

tasks/nc_installation.yml
```diff
@@ -175,3 +175,32 @@
...
+    - name: "[NC apps] - Disable apps"
+      ansible.builtin.command: php occ app:disable "{{ item }}"
+      become_user: "{{ nextcloud_websrv_user }}"
+      become_flags: "{{ ansible_become_flags | default(omit) }}"
+      become: true
+      register: _occ_app_disable_result
+      changed_when:
+        - "'disabled' in _occ_app_disable_result.stdout"
+        - "'No such app' not in _occ_app_disable_result.stdout"
+      with_items: "{{ nextcloud_disable_apps }}"
```
aalaesar, line 206: My last suggestion was incomplete.
the block can be dropped and the indentation for the simple task remaining must be adjusted.
I recommand testing this step in CI by disabling some apps in the molecule tests
1 that is never shipped (apporder ?) and 1 that is always shipped (photo ?)

```suggestion
  - name: "[NC apps] - Disable Nextcloud apps"
    ansible.builtin.command: php occ app:disable "{{ item }}"
    become_user: "{{ nextcloud_websrv_user }}"
    become_flags: "{{ ansible_become_flags | default(omit) }}"
    become: true
    register: _occ_app_disable_result
    changed_when:
      - "'disabled' in _occ_app_disable_result.stdout"
      - "'No such app' not in _occ_app_disable_result.stdout"
    with_items: "{{ nextcloud_disable_apps }}"
```

aalaesar (dismissed):

staticdev:
@jspaans91 friendly ping on the example ;D I want to merge this one.

tasks/nc_installation.yml
```diff
@@ -175,3 +175,14 @@
     - config
     - themes
     - updater
+
+- name: "[NC apps] - Disable Nextcloud apps"
```
wiktor2200, line 179: @staticdev What do you think about moving this whole step to [nc_apps.yml](https://github.com/aalaesar/install_nextcloud/blob/139ec82d1d7a2df7f4c9a5d93b7b57c7d32cdf9c/tasks/nc_apps.yml) to have all app management within one file?

aalaesar: Hello.
@staticdev @wiktor2200
I think it's good as is for one last release for 1.x
It'll be reworked on v2 with a module to manage apps.
Probably one of the first modules I'll work on.

aalaesar: Still need a Readme example and some CI tests 😀

staticdev: @aalaesar maybe we can drop this one an refactor when we have the occ module from https://github.com/aalaesar/install_nextcloud/pull/198

aalaesar: @staticdev hm. We are just missing some test and doc. Let me finish this today so we'll merge it Before 2.0. Idealy a manage app module would bring more  value

aalaesar:
Should be all that was missing.
can you review  @staticdev  ?
I've commited to much to properly review

aalaesar:
ho @staticdev just realised that tests playbook  under /tests/ are not used anymore. I am right ?
but that mean that code coverage ha reduced. 😞

molecule/default/latest-converge.yml
```diff
@@ -10,6 +10,8 @@
         cache_valid_time: 600
       when: ansible_os_family == 'Debian'
       changed_when: false
-
+  vars:
+    nextcloud_disable_apps:
+      - photos
```
wiktor2200, line 15: Could you add 2 apps to test scenario? Sometimes lists behave a little different for multiple values, so we would catch this misbehavior.

tests/nextcloud_mysql.yml
```diff
@@ -25,3 +25,5 @@
     nextcloud_apps:
       files_external: ""  # enable files_external which is already installed in nextcloud
       calendar: "https://github.com/nextcloud/calendar/releases/download/v1.5.0/calendar.tar.gz"  # download and install calendar app
+    nextcloud_disable_apps:
```
wiktor2200, line 28: I think these tests are not used any more. These are some obsolete Vagrant files.

wiktor2200 (changes requested):

molecule/default/latest-converge.yml
aalaesar: I've added theming and mail in my latest commit

tests/nextcloud_mysql.yml
aalaesar: @staticdev 
Yes they're not , but they're not for vagrant. I wrote them when I started the role and was using circleCI.
They seems to be good valid playbooks that could help with code coverage.

PS: I'm thinking about a way to do better code  coverage in our current molecule that would not take ages to do.
I'll have to do some tests with the `import_role` module and the `tasks_from` parameter for ad-hoc unit tests of each tasks file.
or maybe I'm reinventing the wheel and a project for doing that exists already.

wiktor2200: So maybe it's worth adding them to molecule. Obviously not in this PR, but maybe as a next step to test more use cases.

wiktor2200 (approved):

staticdev:
Lgtm @aalaesar ?

PR #189, opened 2023-01-07
fix(occ config:system:set): Use --type=boolean switch

### LINKED ISSUE #187: nextcloud_config_settings: Let 'php occ config:system:set' make use of switch '--type=boolean'
In the current release individual configuration options including bool ones are ment to be set like this:

```
nextcloud_config_settings:
  - { name: 'mail_smtpauth', value: 'false' }
```
But this way the boolean value will be interpreted as string by the occ command. The configured value will not be respected by Nextcloud.

So instead of
`ansible.builtin.command: php occ config:system:set {{ item.name }} --value="{{ item.value }}"`

a better solution might be:
`ansible.builtin.command: php occ config:system:set {{ item.name }} --value="{{ item.value }}" {% if item.value | type_debug == 'bool' %}--type=boolean{% endif %}`

The usage would then be:
```
nextcloud_config_settings:
  - { name: 'mail_smtpauth', value: false }
```
^-- mind the missing quotes not surrounding the bool value anymore.

stevenengland:
To enable correct config settings, the --type=boolean switch for the php
occ config:system:set command was introduced

For fine grained tests, a playbook based on molecule verifier (ansible) was introduced.

Closes #187


staticdev (changes requested):
@stevenengland thanks a lot for your contribution. There are soms minor linting issues. You can check on github changes as they are highlighted so we can merge.

stevenengland:
> @stevenengland thanks a lot for your contribution. There are soms minor linting issues. You can check on github changes as they are highlighted so we can merge.

Hi @staticdev , As this is my first time contribution to a public repo and an ansible role in particular I am facing some sort of startup issues here :-D Like the fact that the linting issues are not thrown when I run the molecule test command locally :) If I manage to get this PR fully functional, I wil submit a second PR regarding the linting thingi... ;)

molecule/default/verify.yml
```diff
@@ -0,0 +1,15 @@
...
+    # load configurations references
+    os_config_ref: "{{ lookup('ansible.builtin.template', '../../defaults/os_config_ref.yml') | from_yaml }}"
+    php_config_ref: "{{ lookup('ansible.builtin.template', '../../defaults/php_config_ref.yml') | from_yaml }}"
+  vars_files:
+    - ../../defaults/main.yml
+
+  tasks:
+    - name: "Include test_install_nextcloud"
+      ansible.legacy.include_role:
+        name: "test_install_nextcloud"
```
staticdev, line 15: Here should be `ansible.builtin.import_role`

staticdev (changes requested):
This is great, but don't worry it happens to everyone even when you do it for 10 yrs some repos can have some specific way to work. I just found just a last detail.

molecule/default/verify.yml
aalaesar, line 15: > Here should be `ansible.builtin.import_role`

here's the suggestion for direct commit if you wish

```suggestion
    - name: "Include test_install_nextcloud"
      ansible.builtin.import_role:
        name: "test_install_nextcloud"
```

aalaesar (changes requested):
Once @staticdev comment is fixed, LGTM
Thank you @stevenengland for your contribution and congratulation. 🥇

molecule/default/verify.yml
stevenengland: Done with the help of @aalaesar 's suggestion :)

aalaesar (approved):
LGTM

staticdev (approved):
Lgtm

PR #190, opened 2023-01-10
test(lint): Add lint step to molecule.yml

stevenengland:
This lint step enables tests to fail locally before submitting code to
the remote repo where github actions will fail because of linting errors.
---
On the local machine yamllint as well as ansible-lint need to be installed. Like so:
python3 -m pip install molecule[lint,<more packages>] ansible-lint

### DIFF
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -19,7 +19,7 @@ provisioner:
     converge: "${MOLECULE_NC:-latest}-converge.yml"
 verifier:
   name: ansible
-#lint: |
-#  set -e
-#  yamllint .
-#  ansible-lint
+lint: |
+  set -e
+  yamllint .
+  ansible-lint
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -19,10 +19,3 @@ provisioner:
     converge: "${MOLECULE_NC:-latest}-converge.yml"
 verifier:
   name: ansible
-lint: |
-  set -e
-  echo "Molecule linting requested: ${TEST_RUN_MOLECULE_LINT}"
-  if [ -z "${TEST_RUN_MOLECULE_LINT}" ] || [ "${TEST_RUN_MOLECULE_LINT}" = true  ]; then
-    yamllint .
-    ansible-lint
-  fi
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -21,5 +21,8 @@ verifier:
   name: ansible
 lint: |
   set -e
-  yamllint .
-  ansible-lint
+  echo "Molecule linting requested: ${TEST_RUN_MOLECULE_LINT}"
+  if [ -z "${TEST_RUN_MOLECULE_LINT}" ] || [ "${TEST_RUN_MOLECULE_LINT}" = true  ]; then
+    yamllint .
+    ansible-lint
+  fi
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,7 +44,7 @@ jobs:
 
       - name: Install test dependencies
         run: |
-          python3 -m pip install --constraint=.github/workflows/constraints.txt ansible molecule[lint,docker] docker netaddr ansible-lint
+          python3 -m pip install --constraint=.github/workflows/constraints.txt ansible molecule[docker] docker netaddr
 
       # See: https://github.com/geerlingguy/ansible-role-mysql/issues/422
       - name: Disable AppArmor (only Debian 10).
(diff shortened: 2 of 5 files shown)


wiktor2200:
Hello @stevenengland! Adding linter to molecule is unnecessary. YAML Linter and Ansible Linter are added to Github Actions as separate steps. Adding linker to molecule is extending running time, while having 4 test cases as we do, it's extending it in every Molecule run.

The best idea is to install and run `ansible-lint` on local machine or just use Github Actions version. 

I've also have a good tip, when you push to your branch and you are trying to make PR, you have this branch available on your profile in your fork and Ansible lint is run as Action anyway but only in your workspace: https://github.com/stevenengland/ansible_nextcloud/actions , so you can use it from there before adding PR :)

wiktor2200 (changes requested):
Adding ansible-lint to molecule is unnecessary

stevenengland:
Hi @wiktor2200, thank you for your explaination. The way you write it, I now understand the caveat very well. That makes the solution I wanted to introduce less than ideal. Since I'm stuck with the idea that the local test pipeline doesn't automatically include linting for everyone contributing, I'm wondering if an alternative would be useful that only takes effect for local tests (when running molecule lint/test) but doesn't run when the code is checked in online, where the Github Actions check linting. You think that this is possible? E.g. by checking certain Github environment variables for presence and if they are not then run the linter locally. I will have a look into this and then you can vote for or against that solution :)

staticdev:
> Hi @wiktor2200, thank you for your explaination. The way you write it, I now understand the caveat very well. That makes the solution I wanted to introduce less than ideal. Since I'm stuck with the idea that the local test pipeline doesn't automatically include linting for everyone contributing, I'm wondering if an alternative would be useful that only takes effect for local tests (when running molecule lint/test) but doesn't run when the code is checked in online, where the Github Actions check linting. You think that this is possible? E.g. by checking certain Github environment variables for presence and if they are not then run the linter locally. I will have a look into this and then you can vote for or against that solution :)
> 
> Update: When I have a look to other repos where workflows fail, it is often the linting :-D Maybe because of forgotten or differently configured local linting ;)

@stevenengland good point.. The way to do it locally is through pre-commit.

molecule/default/molecule.yml
```diff
@@ -19,7 +19,7 @@ provisioner:
     converge: "${MOLECULE_NC:-latest}-converge.yml"
 verifier:
   name: ansible
-#lint: |
```
staticdev, line 22: In fact would be nice to delete these commented lines.

staticdev (changes requested):

stevenengland:
> 

Wow, what a powerful tool. I really like the idea of the pre-commit project. I will test it. Just for further reference and to teach myself I finished the idea of using a environment variable approach to distinguish between local and remote tests.

What do you say: Shall I instead refactor the code to use a pre-commit config or shall the complete PR/idea of local forced linting be rejected?

staticdev:
> 

Actually, if you add pre-commit config you will be achieving both, since when you have it, you cannot commit if linting does not pass. Basically you just need to copy paste this file to the root and add to our contributing documentation that you need to run: `pre-commit install` to work with this project.

wiktor2200:
`pre-commit` is extremely useful tool, especially in very complex environments with multiple testing schemas or lots of dependencies (such as in js/ts/py projects) or when CI run lasts long time.

But in my opinion using pre-commit is an overkill here. It's additional step before contributing and it requires some dependencies to be installed. I usually commit from multiple devices which has different configurations or sometimes when small changes are needed even from browser via Github Editor. So now I only need my Github account or SSH/GPG keys to contribute and it can be done from anywhere with clean terminal, `pre-commit` limits that.

@aalaesar WDYT about it?

stevenengland:
> `pre-commit` is extremely useful tool, especially in very complex environments with multiple testing schemas or lots of dependencies (such as in js/ts/py projects) or when CI run lasts long time.
> 
> But in my opinion using pre-commit is an overkill here. It's additional step before contributing and it requires some dependencies to be installed. I usually commit from multiple devices which has different configurations or sometimes when small changes are needed even from browser via Github Editor. So now I only need my Github account or SSH/GPG keys to contribute and it can be done from anywhere with clean terminal, `pre-commit` limits that.
> 
> @aalaesar WDYT about it?

Also a very interesting opinion. Thanks for sharing. From my point of view I am unsure. I made the refactoring, now you can decide what is the best solution :)

staticdev:
> 

@wiktor2200 actually this is not the case. pre-commit is only enforced when you do a pre-commit install, it will not limit you from doing any changes from Github directly actually. I have pre-commit in all my projects also, it is really helpful I would say. No downsides except Dependabot does not update versions on pre-commit config.

staticdev (approved):
LGTM.

wiktor2200 (approved):
LGTM

PR #193, opened 2023-01-12
remove tag always

### LINKED ISSUE #191: Remove the `always` tag from the role.
Hello,

First of all, thanks for this role, is has been very useful to us at [La Quadrature Du Net](https://lqdn.fr/nous).

I'm opening this issue to ask if I can make a pull request to remove the `always` tag that is being used in this role. 

I figure it is not necessary to have such a tag on a role. It induces issues when running parts of a playbook with many roles on a host with this role, as it might run part of this role when it is not needed, or when it is needed that it _must not_ run. 

I know I can skip the tasks altogether with the `--skip-tags always` option, but [per the documentation](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_tags.html#special-tags-always-and-never), it is bad practice as it will also skip fact gathering.

Overall, why does this role need the `always` tag ? If it does not, can I remove it & request a pull to apply this change in the futur and avoid a fork ?

Thanks again, 

nono

aalaesar:
change it to the only tag present in the role
fix #191

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -10,6 +10,7 @@
   ansible.builtin.include_tasks: php_install.yml
   tags:
     - install_apps
+
 - name: Install web server packages
   when: nextcloud_install_websrv
   ansible.builtin.include_tasks: ./websrv_install.yml
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -3,12 +3,12 @@
 
 - name: "Load os specific variables"
   ansible.builtin.include_tasks: ./setup_env.yml
-  tags: always
-
+  tags:
+    - install_apps
 - name: "Install PHP packages"
   ansible.builtin.include_tasks: php_install.yml
-  tags: always
-
+  tags:
+    - install_apps
 - name: Install web server packages
   when: nextcloud_install_websrv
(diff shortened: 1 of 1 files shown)


tasks/main.yml
```diff
@@ -3,12 +3,12 @@
 
 - name: "Load os specific variables"
   ansible.builtin.include_tasks: ./setup_env.yml
-  tags: always
-
+  tags:
+    - install_apps
```
wiktor2200, line 7:
```suggestion
    - install_apps

```

tasks/main.yml
```diff
@@ -3,12 +3,12 @@
...
 - name: "Install PHP packages"
   ansible.builtin.include_tasks: php_install.yml
-  tags: always
-
+  tags:
+    - install_apps
```
wiktor2200, line 11:
```suggestion
    - install_apps

```

wiktor2200 (changes requested):
I've added suggestions to apply with missing newlines.

nono-lqdn (approved):
seems good ! Thanks @aalaesar and @wiktor2200

wiktor2200 (approved):

PR #195, opened 2023-01-20
Make nextcloud_websrv variable consistent in whole repo

wiktor2200:
Hi! While doing code review I've noticed, that there are some inconsistencies in variable usage. So I've fixed it.
* https://github.com/aalaesar/install_nextcloud/blob/2edb1a7e8b0b6b9fb1034a41d5fc6f25a6895346/tasks/main.yml#L38
* https://github.com/aalaesar/install_nextcloud/blob/2edb1a7e8b0b6b9fb1034a41d5fc6f25a6895346/tasks/main.yml#L42

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -39,7 +39,7 @@
 
     - name: Configure Apache web server
       ansible.builtin.include_tasks: ./http_apache.yml
-      when: nextcloud_websrv in ["apache", "apache2"]
+      when: nextcloud_websrv in ["apache2"]
 
 - name: Configure Redis server
   ansible.builtin.include_tasks: ./redis_server.yml
--- a/tasks/websrv_install.yml
+++ b/tasks/websrv_install.yml
@@ -10,7 +10,7 @@
   ansible.builtin.package:
     name: "libapache2-mod-php{{ php_ver }}"
     state: present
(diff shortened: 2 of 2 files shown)


staticdev:
@wiktor2200 as far as I understand it is correct how it is today and also consistent with the documentation. nextcloud_websrv cannot be apache, just apache2. Can you explain? Maybe I am missing something.

wiktor2200:
`["apache", "apache2"] ` was always there in line 42. 
`apache2` is available since forever 2002, but sometimes people use shortcuts. So it will cover such shortcuts. It's like fail safe switch I guess.
Anyway I don't push on changing it, I just wanted it to be consistent. So if you would prefer removing `apache` here and leave `apache2` I will change ;) Up to you :)

aalaesar:
found the commit that introduced this apache/Apache2 alias.
It was a retrocompatible change I made 6 years ago : https://github.com/aalaesar/install_nextcloud/commit/ab1acaf8e2b0dcfba20d707b477b85460ddc4d48

I don't think `apache `is much used nowadays but I don't have any objection about making the code consistent, one way or another.

wiktor2200:
So I'm leaving only `apache2` and `nginx` options as it's documented in README.

staticdev (approved):
LGTM.

PR #198, opened 2023-01-21
Add run occ module

aalaesar:
Branch new module for running raw occ commands.
Also prepare foundation code for all other modules using occ


aalaesar:
oops. still have to change/remove Licensing headers

plugins/module_utils/occ.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2022, Marc Crébassa <aalaesar@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+import os
+# import json
```
staticdev, line 7: Better to not have commented code on main branch.

plugins/module_utils/occ.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def run_occ(module, command, ):
```
staticdev, line 10: Formatting of this file is very odd. Try to run black in whole folder, also useful to have on pre-commit.

[.pre-commit-config.yaml](https://github.com/psf/black/blob/main/.pre-commit-config.yaml):

```
---
repos:
  - repo: https://github.com/psf/black
    rev: 22.12.0
    hooks:
      - id: black
        language_version: python3.11
```

staticdev (changes requested):
Some suggestions for QA.

plugins/module_utils/occ.py
aalaesar: cleaned up  the commented line

plugins/module_utils/occ.py
aalaesar: just done it.
how about now ? 😃

aalaesar:
I'll try go see if I can add some basic [unit testing ](https://docs.ansible.com/ansible/latest/dev_guide/testing_units_modules.html) in the next couple of days

staticdev (approved):
LGTM. We have a basis to add tests and other code after that.

PR #199, opened 2023-01-22
Cleanup unnecessary php module params

staticdev:
* For our target OSs it will always be "php{{ php_ver }}-apcu" like the rest of packages on https://github.com/aalaesar/install_nextcloud/blob/main/tasks/php_install.yml#L16
* Fix typo of redix instead of redis
* php{{ php_ver }}-redis is the needed dependency for redis, simplified by removing from php_pkg_spe


aalaesar (approved):
LGTM

PR #203, opened 2023-02-02
Complete collection metadatas

aalaesar:


aalaesar:
Apparently ansible-lint request a changelog file. maybe we could generate it/ append it with the release drafter ?  WDYT @staticdev ?
or maybe an other CI tool if it can't do it.

wiktor2200:
It may be harder for old releases (broken and removed when fixing galaxy). But there is this fancy tool: https://github.com/github-changelog-generator/github-changelog-generator which helps generating Changelog.md file from git tags, issues and merged pull requests.
And there is even github action for that ;) https://github.com/marketplace/actions/generate-changelog

aalaesar:
Hm this changelog file is kind of a constraint in a CI flow. It imply that something/one must commit on it just when doing release. 
So how can this work ? we make a PR with a special case or tag for releasing and the ci make a commit that update the Release file in the branch only once ? Or simply a github action or a local tool (like pre-commit) that can be triggered manually ?

staticdev:
@aalaesar personally I wouldn't care much about polder versions since collection is basically a new project. If you think this way life is simpler and we don't have a problem. Also for old incompatible tags, maybe easy solution is just delete. I don't see a good use case to install a version from 1 year ago anymore.

aalaesar:
> @aalaesar personally I wouldn't care much about polder versions since collection is basically a new project. If you think this way life is simpler and we don't have a problem. Also for old incompatible tags, maybe easy solution is just delete. I don't see a good use case to install a version from 1 year ago anymore.

Indeed @staticdev, that's a good idea for when the repo is handled to nextcloud to reset the version counter to 1.0.0
That's a detail. I'm focusing on automating updating the changelog file. or maybe we just put a link to the github release page, so we avoid to change the CI at all. would it be good practice ?

staticdev:
> > @aalaesar personally I wouldn't care much about polder versions since collection is basically a new project. If you think this way life is simpler and we don't have a problem. Also for old incompatible tags, maybe easy solution is just delete. I don't see a good use case to install a version from 1 year ago anymore.
> 
> Indeed @staticdev, that's a good idea for when the repo is handled to nextcloud to reset the version counter to 1.0.0 That's a detail. I'm focusing on automating updating the changelog file. or maybe we just put a link to the github release page, so we avoid to change the CI at all. would it be good practice ?

I like the idea of the link. Just keep it simpele then we use the energy to nextcloud features and security.

wiktor2200:
I like idea with the link too. It seems to be the easiest.

aalaesar:
@staticdev @wiktor2200  
simplified the changelog file to link to the release page.

staticdev (approved):
Lgtm

README.md
```diff
@@ -0,0 +1,120 @@
...
+This collection has been tested against following Ansible versions: **>=2.10.0**.
+
+Plugins and modules within a collection may be tested with only specific Ansible versions.
+<!--end requires_ansible-->
+
+## Python Support
+
+* Collection tested on 3.10+
+
+Note: Python2 is deprecated from [1st January 2020](https://www.python.org/doc/sunset-python-2/). Please switch to Python3.
```
staticdev, line 19: @aalaesar i would remove note over Py2 in 2023 actually.

aalaesar: Good finding ! Just removed it

Issue #207, opened 2023-02-09
ansible-lint behaving differently between CI and installed pre-commit hooks


aalaesar:
Hello there !
I actually can't commit locally with pre-commit installed because of ansible-lint errors.
That could be a normal behaviour **if it was'nt errors about already commited code** that did not triggered our CI tests.
I've looked at https://github.com/ansible/ansible-lint-action to see if there were any config in particular but found nothing, as expected because that's how galaxy is running it.

I forced installed the same ansible-lint version used in the release for 6.0.0 to 6.11.0 but every time I got the same errors: 

```bash
$ git status
Sur la branche nextcloud-rebranding
Modifications qui seront validées :
  (utilisez "git restore --staged <fichier>..." pour désindexer)
        modifié :         galaxy.yml

$ git commit -S -m "update collection namespace and name"
yamllint.................................................................Passed
Ansible-lint.............................................................Failed
- hook id: ansible-lint
- exit code: 2
[...]
WARNING  Listing 16 violation(s) that are fatal
# I get a bunch of thoses:
no-free-form: Avoid using free-form when calling module actions. (ansible.builtin.command) (warning)
roles/install_nextcloud/tasks/db_mysql.yml:33 Task/Handler: [mySQL] - Get list of hosts for the root user.

# and then the fatal errors : 
name[template]: Jinja templates should only be at the end of 'name'
roles/install_nextcloud/tasks/nc_apps.yml:10 Task/Handler: [App] - Verify the yaml declaration for the app "{{ nc_app_name }}"

name[template]: Jinja templates should only be at the end of 'name'
roles/install_nextcloud/tasks/nc_apps.yml:17 Task/Handler: [App] - Install from the official app-store app: "{{ nc_app_name }}"

name[template]: Jinja templates should only be at the end of 'name'
roles/install_nextcloud/tasks/nc_apps.yml:26 Task/Handler: [App] - Download Archive in apps folder from "{{ nc_app_source }}"

name[template]: Jinja templates should only be at the end of 'name'
roles/install_nextcloud/tasks/nc_apps.yml:38 Task/Handler: [App] - Enable the application "{{ nc_app_name }}"

name[template]: Jinja templates should only be at the end of 'name'
roles/install_nextcloud/tasks/nc_apps.yml:47 Task/Handler: [App] - Configure the application "{{ nc_app_name }}"

You can skip specific rules or tags by adding them to your configuration file:
# .config/ansible-lint.yml
[...]

PR #210, opened 2023-02-11
Add pre-commit update action

wiktor2200:
Hi @aalaesar  @staticdev! I've added github action for pre-commit update to mitigate issues like: #207 
It's based on example from here: https://github.com/browniebroke/pre-commit-autoupdate-action

It requires a token: `token: ${{ secrets.GITHUB_TOKEN }}` for pull request creation, could you help with that?

### DIFF
--- a/.github/precommit-update.yml
+++ b/.github/precommit-update.yml
@@ -1,0 +1,25 @@
+name: Pre-commit auto-update
+
+on:
+  # every day at midnight
+  schedule:
+    - cron: "0 0 * * *"
+
+
+jobs:
+  auto-update:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


staticdev (changes requested):
Thanks @wiktor2200 Just check for annoying lint warnings on gh since they eventually become errors in following lint versions.

About GITHUB_TOKEN it will just work =) other workflows also need it

wiktor2200:
Fixed. I've also noticed that I put it in wrong directory, so I moved it to proper one.

aalaesar (approved):
LGTM

staticdev (approved):

PR #209, opened 2023-02-09
Change namespace to nextcloud and collection name to admin

aalaesar:
requires:
- [x] #206 merged
- [x] #185 and #171 done
- [x] #216  merged
- [x] #186 done

changes : 
- changed all references of:
  - aalaesar.nextcloud => nextcloud.admin
  - aalaesar/install_nextcloud  => nextcloud/**ansible-collection-nextcloud-admin**  (expected repository new name)
- updated some titles

PR #216, opened 2023-02-22
Galaxy deploy collection

### LINKED ISSUE #186: Adjust ansible galaxy to import the collection from the nextcloud namespace
Check how to add/move the collections to the new namespace into galaxy (trying to keep stats) 
- [x] https://galaxy.ansible.com/docs/contributing/namespaces.html

staticdev:
Closes #186


.github/workflows/release.yml
```diff
@@ -44,11 +44,13 @@ jobs:
...
+      - name: Release to Ansible Galaxy.
         run: >-
-          ansible-galaxy role import
-          --api-key ${{ secrets.GALAXY_TOKEN }} $(echo ${{ github.repository }}
-          | cut -d/ -f1) $(echo ${{ github.repository }} | cut -d/ -f2)
+          ansible-playbook -i 'localhost,' galaxy-deploy.yml
+          -e "namespace=$(echo ${{ github.repository }}
+          | cut -d/ -f1)
+          -e "collection"=$(echo ${{ github.repository }}
+          | cut -d/ -f2)
```
aalaesar, line 53: looks like it's missing 
```bash
#   -e "github_tag=${{ github.ref }}"
```
as it is used in various places in the playbooks

also this step was run everytime a merge was done on master.
so It may create issues to push branch master and tags everytime on galaxy.
Shouldn't we just to galaxy only when adding a new tag ?


Well As long as #209 is not merged we should be able to test that in my namespace

galaxy-deploy.yml
```diff
@@ -0,0 +1,50 @@
...
+        content: |
+          token: {{ lookup('env','ANSIBLE_GALAXY_TOKEN') }}
+        dest: ~/.ansible/galaxy_token
+
+  tasks:
+    - name: Ensure the galaxy.yml tag is up to date.
+      lineinfile:
+        path: galaxy.yml
+        regexp: "^version:"
+        line: 'version: "{{ release_tag }}"'
```
aalaesar, line 43: I don't recommend that.
We could put an assert instead that check the coherence between `current-version.txt`, the detected tag by `release_tag` and the version tag in the galaxy.yml file.
maybe we can drop one of the 3

aalaesar (changes requested):
I noted that this workflow is run at every merge into main.
but for the collection we don't want this.
The ansible-collection build the archive with the content as is and use the version metadata to generate the archive name.
This will break/fail on every commit in main that are not tags. ([release_tag won't have the proper value on non tag event](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context))
So we should probably change-the trigger or add-a-new workflow for release on tag only, if possible

.github/workflows/release.yml
staticdev: Good catch, fixed.

galaxy-deploy.yml
staticdev: Ok, what I did is I changed now the command to check changed version, since `galaxy.yml` with version is mandatory we don't need `current-version.txt` anymore.

staticdev:
> I noted that this workflow is run at every merge into main. but for the collection we don't want this. The ansible-collection build the archive with the content as is and use the version metadata to generate the archive name. This will break/fail on every commit in main that are not tags. ([release_tag won't have the proper value on non tag event](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)) So we should probably change-the trigger or add-a-new workflow for release on tag only, if possible

For now I would keep it to check if deployment is ok like this and we can create another issue to improve it if you agree.

galaxy-deploy.yml
aalaesar: Hm. 
So at every merge commit in main, the playbook change version value to main in galaxy.yml. then the build task will generate the archive with main as suffix. This might work  for the build task but for the publish will very probably fail as the version will not be in semver format.
I'll try locally

staticdev: Not sure I understood what you mean. You need to intentionally change the version through your PR, so it will not always be the case of every merge commit, no? What am I missing?

staticdev:
> I noted that this workflow is run at every merge into main. but for the collection we don't want this. The ansible-collection build the archive with the content as is and use the version metadata to generate the archive name. This will break/fail on every commit in main that are not tags. ([release_tag won't have the proper value on non tag event](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)) So we should probably change-the trigger or add-a-new workflow for release on tag only, if possible

I see @aalaesar, maybe this will happen but I expect it to work. I propose we do a couple of tests as it is. If we get problems every merge I will make a PR to check for new tag. My idea is to try to have as much releases as possible to identify and correct errors more easily also.

aalaesar:
For testing you can change the condition to trigger the action for every commit in this branch.
So we'll see right away

staticdev:
> For testing you can change the condition to trigger the action for every commit in this branch. So we'll see right away

Fair enough, I just tried that.. current problem is that on the new repo there must be a secret set for Ansible Galaxy: 

`fatal: [localhost]: FAILED! => {"changed": false, "msg": "A valid ANSIBLE_GALAXY_TOKEN must be set."}`

aalaesar:
> Fair enough, I just tried that.. current problem is that on the new repo there must be a secret set for Ansible Galaxy:
> 
> `fatal: [localhost]: FAILED! => {"changed": false, "msg": "A valid ANSIBLE_GALAXY_TOKEN must be set."}`

The secret is still here but it was named GALAXY_TOKEN when we first needed it. Any way, I'll add it again with the ANSIBLE_ prefix 😄

.github/workflows/release.yml
```diff
@@ -38,17 +38,20 @@ jobs:
...
-            cat current-version.txt
+            cat galaxy.yml|grep version:|cut -d' ' -f2
 
       - name: Install Ansible
         run: |
           pip install --constraint=.github/workflows/constraints.txt ansible
 
-      - name: Trigger a new import on Galaxy.
+      - name: Release to Ansible Galaxy.
         run: >-
```
aalaesar, line 48: Existing token needs to be set into env for the playbook to pass the assert
```suggestion
      - name: Release to Ansible Galaxy.
        env:
          ANSIBLE_GALAXY_TOKEN: ${{ secrets.GALAXY_TOKEN }}
        run: >-
```

aalaesar (changes requested):

.github/workflows/release.yml
```diff
@@ -38,17 +38,22 @@ jobs:
...
-          ansible-galaxy role import
-          --api-key ${{ secrets.GALAXY_TOKEN }} $(echo ${{ github.repository }}
-          | cut -d/ -f1) $(echo ${{ github.repository }} | cut -d/ -f2)
+          ansible-playbook -i 'localhost,' galaxy-deploy.yml
+          -e "github_tag=${{ github.ref }}"
+          -e "namespace=$(echo ${{ github.repository }}
+          | cut -d/ -f1)"
+          -e "collection=$(echo ${{ github.repository }}
+          | cut -d/ -f2)"
```
aalaesar, line 56: looks like we need mode output for debugging
```suggestion
          ansible-playbook -D -vv -i 'localhost,' galaxy-deploy.yml
          -e "github_tag=${{ github.ref }}"
          -e "namespace=$(echo ${{ github.repository }}
          | cut -d/ -f1)"
          -e "collection=$(echo ${{ github.repository }}
          | cut -d/ -f2)"
```

staticdev:
@aalaesar the problem now it that the namespace is diverging on galaxy.yml from one on repo name.

staticdev:
@aalaesar I added hardcoded the old namespace and name to test... Now I get:

`Error: derr": " [ERROR]: Galaxy import error message: Import Task \"25916\" failed: Invalid\ncollection metadata. 'tag' has invalid format: self-hosted\nERROR! Galaxy import process failed: Invalid collection metadata. 'tag' has invalid format: self-hosted (Code: None)",`

.github/workflows/release.yml
```diff
@@ -38,17 +38,23 @@ jobs:
         with:
           tag-template: "{VERSION}"
           version-command: |
-            cat current-version.txt
+            cat galaxy.yml|grep version:|cut -d' ' -f2
```
wiktor2200, line 41: Missing spaces for better yaml read:
```suggestion
            cat galaxy.yml | grep version: | cut -d' ' -f2
```

galaxy-deploy.yml
```diff
@@ -0,0 +1,50 @@
...
+      fail:
+        msg: A valid ANSIBLE_GALAXY_TOKEN must be set.
+      when: "lookup('env','ANSIBLE_GALAXY_TOKEN') | length == 0"
+
+    - name: Ensure the ~/.ansible directory exists.
+      file:
+        path: ~/.ansible
+        state: directory
+
+    - name: Write the Galaxy token to ~/.ansible/galaxy_token
```
wiktor2200, line 32: Could you confirm that this token won't be shown in any pipeline log?

meta/runtime.yml
```diff
@@ -0,0 +1,2 @@
+---
+requires_ansible: '>=2.10.0'
```
wiktor2200, line 2: Ansible lint is complaining about this line.
```
meta-unsupported-ansible
requires_ansible key must be set to a supported version.
```

staticdev:
@aalaesar @wiktor2200 TL;DR it works. Now fixing the errors from ansible-lint, there are a few.

meta/runtime.yml
staticdev: i fixed it. Neede to be 2.11.0.

galaxy-deploy.yml
aalaesar: Yes maybe put a no_log here to be sure

aalaesar:
Hello there 
Well done @staticdev !
I think I might  not understand completely how the code behaviour, but [it worked](https://galaxy.ansible.com/search?deprecated=false&keywords=Nextcloud&order_by=-relevance&page=1) ! ( But ansible page display 0 modules...)

So if the namespace and name are now hard coded in the playbook command, we should set the version to 2.0.0 and merge it after #209, once the namespace has been created.

.github/workflows/release.yml
```diff
@@ -38,17 +38,22 @@ jobs:
...
+      - name: Release to Ansible Galaxy.
+        if: "steps.check-version.outputs.tag"
+        env:
+          ANSIBLE_GALAXY_TOKEN: ${{ secrets.GALAXY_TOKEN }}
         run: >-
-          ansible-galaxy role import
-          --api-key ${{ secrets.GALAXY_TOKEN }} $(echo ${{ github.repository }}
-          | cut -d/ -f1) $(echo ${{ github.repository }} | cut -d/ -f2)
+          ansible-playbook -i 'localhost,' galaxy-deploy.yml
+          -e "github_tag=$(cat galaxy.yml|grep version:|cut -d' ' -f2)"
```
wiktor2200, line 53: I missed it here, same as above.
```suggestion
          -e "github_tag=$(cat galaxy.yml | grep version: | cut -d' ' -f2)"
```

wiktor2200:
Great work @staticdev! I've left one more small comment, but apart from that LGTM and can be merged. :)
Regarding Galaxy collection, it's weird, maybe counter would be aggregated later, because when you click on `Content` button you can see 1 role, 1 module and 3 plugins, so all what expected.
screen

.github/workflows/release.yml
staticdev: Done.

galaxy-deploy.yml
staticdev: I did some tests yesterday and today, copy does not output data on -vvv. But I added `no_log` anyway to be on safe side.

staticdev:
@aalaesar LGTM?

aalaesar:
> @aalaesar LGTM?

The tests have messed up a little bit the tags and the release (no pb it's fixable).
I still want the next release to be 2.0.0 as the collection is a breaking change
When I'll delete the tags 1.9.x, will the release drafter regenerate the release description from scratch or do I have to put back the current text in the draft ?

staticdev:
> > @aalaesar LGTM?
> 
> The tests have messed up a little bit the tags and the release (no pb it's fixable). I still want the next release to be 2.0.0 as the collection is a breaking change When I'll delete the tags 1.9.x, will the release drafter regenerate the release description from scratch or do I have to put back the current text in the draft ?

Ah, Yeah I messed up. Not drafter fault, I did a couple of releases. I will do a clean up but we will need to manually the description to 2.0.0.. One time thing..

wiktor2200:
Hi @staticdev @aalaesar!
That's not a problem at all :)
My RSS reader has it cached, so I pasted "1.9.0" description to draft :)

galaxy.yml
```diff
@@ -9,7 +9,7 @@ namespace: aalaesar
 name: nextcloud
 
 # The version of the collection. Must be compatible with semantic versioning
-version: 2.0.0
+version: 1.9.3
```
aalaesar, line 12: I'm OK for merging now but the version will trigger a release again .
given the CI (that I still don't fully understand :p), if we put `1.8.0` here, will it skip the release ?

staticdev: Yes @aalaesar, I reverted to 1.8.0, so it will not try to do anything.

aalaesar (approved):
LGTM

staticdev:
> 

I see LGTM from @wiktor2200 also so doing it.

PR #230, opened 2023-03-09
Fix pre-commit update action

wiktor2200:
This should change pre-commit updater commiter to our dependabot.


staticdev:
Let's try.

staticdev (approved):

wiktor2200:
We'll see with next pre-commit update. :)

Issue #234, opened 2023-03-11
Change default GITHUB_TOKEN to generated personal access token for dependabot.


wiktor2200:
Hi @aalaesar, @staticdev!
We need to change default token to custom in pre-commit update here:
https://github.com/nextcloud/ansible-collection-nextcloud-admin/blob/5286bef6116da18a3521c9b8b966b51aec5e6dab/.github/workflows/precommit-update.yml#L23

I don't have sufficient permissions to generate new token for `dependabot[bot]` user. @aalaesar Could you create new token and call it for example: `GITHUB_TOKEN_DEPENDABOT` or `GITHUB_TOKEN_PRECOMMIT_UPDATE`? I'll change above line then.

Longer story with whole problem explained is covered here: https://github.com/peter-evans/create-pull-request/issues/1715

staticdev:
@wiktor2200 I also dont have permission, I already reverted your previous PR

aalaesar:
Ok so I created the Token TOKEN_DEPENDABOT (can't prefix secrets with github) and now the CI commit will be authored by "me" if I'm understanding it correctly.

wiktor2200:
Hi @aalaesar!
Is this token **your** personal token? Or repository specific one?
I think, it's not a good idea if it's yours and we should probably leave it as it is with default `github-action`, when `dependabot` user is not possible here.

The main problem was to differentiate dependencies updated by `dependabot` and having all actions (`yamllint` and `ansible-lint`) run within pre-commit update PR.

But if it's not possible I think we should drop it for know. Versions update are not a big deal and should not break `yamlint` or `ansible-lint` results (especially for pre-commit).

WDYT? @aalaesar @staticdev

aalaesar:
> Hi @aalaesar! Is this token **your** personal token? Or repository specific one? I think, it's not a good idea if it's yours and we should probably leave it as it is with default `github-action`, when `dependabot` user is not possible here.

Yes, It's _just a_ personal acces token , as specified by the action's author. With just enough scope to make the action working.


> But if it's not possible I think we should drop it for know. Versions update are not a big deal and should not break `yamlint` or `ansible-lint` results (especially for pre-commit).
> 
> WDYT? @aalaesar @staticdev
 Could be _nice_ to get matching versions for precommits and tests in ci. But why needing absolutely the ci to trigger linters for those commits ?
 I think the pre-commit  config may be ignored by the linters anyway.

[...]

PR #236, opened 2023-03-16
Change pre-commit update flow

wiktor2200:
Regarding Issue #234

### DIFF
--- a/.github/workflows/precommit-update.yml
+++ b/.github/workflows/precommit-update.yml
@@ -5,8 +5,6 @@ name: Pre-commit auto-update
   # every day at midnight
   schedule:
     - cron: "0 0 * * *"
-  # on demand
-  workflow_dispatch:
 
 jobs:
   pre-commit-auto-update:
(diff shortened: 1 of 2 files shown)


wiktor2200:
Hi @staticdev @aalaesar!
I've seen that there was an update in `apparmor` repo. 
But official Debian10 in Gitlab Actions wasn't updated for some reason. I've added `apt-get update` command to fix that.

BTW, this DCO check is kind of annoying. I've GPG signed commits turned on by default (set up by `git config`), but unfortunately there is no option to do the same for sign-off. 
Do you have any suggestions/ideas how we can improve that? Because even simple `--amend` re-triggers pipelines, so they are run in vain.

aalaesar:
> Hi @staticdev @aalaesar! I've seen that there was an update in `apparmor` repo. But official Debian10 in Gitlab Actions wasn't updated for some reason. I've added `apt-get update` command to fix that.

Ok thank you.


> BTW, this DCO check is kind of annoying. I've GPG signed commits turned on by default (set up by `git config`), but unfortunately there is no option to do the same for sign-off. Do you have any suggestions/ideas how we can improve that? Because even simple `--amend` re-triggers pipelines, so they are run in vain.

hmm maybe putting a dependency for the DCO on the tests not sure if possible

aalaesar (approved):
LGTM

PR #245, opened 2023-03-26
Molecule tests - Add testing scenarios for all supported OS and NC versions

wiktor2200:


wiktor2200:
Solving #245

wiktor2200:
I've also added enhancement for molecule tests, which enhancing testing scenarios for all supported OS and NC versions.

wiktor2200:
The only problem that I can see is running 12 jobs for every commit in PR. I'm thinking about changing it to run only latest for current commit, but run all on schedule once a week to make sure that the collection still works for older OS and versions.
WDYT @aalaesar @staticdev ?

aalaesar:
I would say that 'latest' and nc are the same now
so doing CI for 'latest'/current and current v-1 will only set 8 tests instead of 12.
It will be always up to us to update CI @ each new release

wiktor2200:
Yeah, it really depends on us which version do we want to support. On official nextcloud server repo there is EOL schedule: https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule so for today (28th of March) versions 24, 25 and 26 are supported, that was one of main parameters when I was creating this PR ;)

I'm proposing running all 24, 25, 26(as latest) on schedule once a week to check if still works + badges added to README. 
And running only latest on every pull request for new code changes to easier development of new features. And running all on merge to make sure that succeeded PR action didn't break anything (if so, there will be a need for revert - but it's less probable). :)

staticdev:
> What is more, when you look on history of this file: https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule/_history you can notice that, there are 3 or 2 currently supported versions (at the same time). Now there are 3, but just before releasing 26, there were only two 24 and latest (25). So it's quite dynamic :)

Main advantage is that they are able to keep pace with a small company and deliver inmovation with short release support. I would say it is adequate also for security reasons. Nextcloud users need to be constantly upgrading.

wiktor2200:
Yeah, totaly agree on that. So the question is, do you agree for my proposal @aalaesar @staticdev for these test scenarios. If so I'll modify my PR :)

wiktor2200:
Hi @aalaesar @staticdev! I've pushed my proposed changes, please take a look and review.
I've also changed schedule for `pre-commit update` action to be run once a week instead of daily, because it's quite a spam recently.

staticdev (approved):

staticdev:
Looking good.

wiktor2200:
Hi @aalaesar, could you also approve if you agree with changes? :)

aalaesar:
> Hi @aalaesar, could you also approve if you agree with changes? :)

hello there !
I haven't be able to write a proper answer for the last few days.
it goes down to:
- being worried that the repo may hit an upper limit of free tier github action with too many long jobs running, maybe we could keep the short linter jobs automatic while setting the long tests for on demand run.
- otherwise PR looks good. 👍

aalaesar (approved):
LGTM

staticdev:
@aalaesar I don't think we will reach. Other projects from nextcloud suffer from similar issue and normally they have much more jobs. Also since they have short release cycle, soon we will have to deprecate some older version.

wiktor2200:
Hi! @aalaesar @staticdev thanks for your reviews and merging ;)

@aalaesar that what I was thinking about, but it's unlikely that we would hit 2000 minutes per month from free tier :)
I've checked lasts run of job and for latest sum of minutes run was 31,9min, running scheduled job would take something around: 105min.

All scenario would be run at least 4 times a month (scheduled) so it's ~400mins, then running once per merging PR, so it really depends on us :writing_hand: and dependency updates.

I would suggest looking on amount of minutes left in pipelines if we would hit this limit or be close the easier solution will be turning off full tests on `main` merge.
So full tests would be run only on schedule then and on every commit in PR and merge would be only latest. :)

PR #253, opened 2023-04-10
Fix an issue with builtin.assemble, which does not take account of th…

marioqxx:
…e order of concatentation.

### DIFF
--- a/roles/install_nextcloud/tasks/tls_signed.yml
+++ b/roles/install_nextcloud/tasks/tls_signed.yml
@@ -12,7 +12,7 @@
     nextcloud_tls_chain_file: "{{ nextcloud_tls_chain_file | default(cert_path + nextcloud_instance_name + \".pem\") }}"
   when: nextcloud_tls_src_chain is defined
 
-- name: "[SIGNED TLS] - Certificate is copied to the host"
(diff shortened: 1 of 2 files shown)


roles/install_nextcloud/tasks/tls_signed.yml
```diff
@@ -36,18 +36,19 @@
...
-    mode: 0640
+    mode: "0640"
     force: false
   when: nextcloud_tls_src_chain is defined
 
-- name: "[SIGNED TLS] - Create chained certificate file for nginx"
-  ansible.builtin.assemble:
-    src: "{{ cert_path }}"
-    regexp: "{{ nextcloud_instance_name }}\\.crt|{{ nextcloud_instance_name }}\\.pem"
+- name: "[SIGNED TLS] - Overwrite and create chained certificate file for nginx"
```
aalaesar, line 43: This seems to break idempotence as the file was also copied a few tasks earlier on the remote host.

marioqxx: Thanks @aalaesar. My new commit shall fix this. The changes are minor, but I did re-arrange the tasks, which shall improve reading the code.

wiktor2200 (approved):
LGTM

staticdev (approved):

PR #254, opened 2023-04-13
Bump version 2.0.1

staticdev:


wiktor2200 (approved):
LGTM. What do you think about merging this #253 first before bumping version? 
There would be all known bugs fixed.

PR #259, opened 2023-04-21
New module: app_info

aalaesar:
Add a simple module to get application informations (status, update, etc)
This is intended to be a foundation for the next module for app management


staticdev:
@aalaesar package managers in general resolve better if you send the complete list of apps in one command. I know this works for example for apt and pip packages. Not sure if it is the case for occ, but could be worth to try for make it more efficient.

aalaesar:
> @aalaesar package managers in general resolve better if you send the complete list of apps in one command. I know this works for example for apt and pip packages. Not sure if it is the case for occ, but could be worth to try for make it more efficient.

hmm
I suspected that calling occ would be costly in time (maybe that's how the run_occ function is implemented that is not optimized)
There is indeed the risk for the module to badly scale (in running time) in case of large amount of application installed on an instances.
In that case I think splitting the information gathering In Two modules would be simplier:
- one module to get **facts** about the nexcloud server with general information (all applications and update available), easy to get in a static and limited number of calls to occ.
- one app_info module to get more details about one application in particular

aalaesar:
Hello there.
I'm back after some personal issues that took me away from my projects.
I wanted to finish the app_info module.
From my point of vue, the module is finished
but the CI seems broken with concurrency issues with docker and the tests running in parallel, and also the occ install command failing somehow 😕
as the change is not related to the installation role, this can be merged.

aalaesar:
@staticdev @wiktor2200 
Gentle up :smile:  could you review the module?
it's a fairly simple one as it only gather infos for only one app

wiktor2200 (approved):
Hi! @aalaesar sorry for late review.
I've analyzed your code and it looks good. 
I'm not convinced only by naming for "shipped" apps, I'd personally rename it to "pre-installed", "default", "enabled_by_default' or similar.
Here in documentation https://docs.nextcloud.com/server/latest/admin_manual/apps_management.html#apps-management they call them: `default enabled` or `enabled by default`.

But it's up to you, so I leave approve for your code :)

plugins/modules/app_info.py
```diff
@@ -0,0 +1,194 @@
...
+      returned: success
+    app_path:
+      description: The full path to the application folder.
+      type: string
+      returned: success
+"""
+
+import copy
+import json
+from ansible.module_utils.basic import AnsibleModule
```
staticdev, line 102: It is normally good to separate build-in imports from 3rd party

```
import json

from ansible.module_utils.basic import AnsibleModule
```

plugins/modules/app_info.py
```diff
@@ -0,0 +1,194 @@
...
+from ansible_collections.nextcloud.admin.plugins.module_utils.occ import run_occ
+from ansible_collections.nextcloud.admin.plugins.module_utils.occ_args_common import (
+    OCC_ARGS_SPEC,
+)
```
staticdev, line 106: I would do long imports like this:

```
from ansible_collections.nextcloud.admin.plugins.module_utils.occ import \
    run_occ
from ansible_collections.nextcloud.admin.plugins.module_utils.occ_args_common import \
    OCC_ARGS_SPEC
```

plugins/modules/app_info.py
```diff
@@ -0,0 +1,194 @@
...
+      description: What is the version proposed for update.
+      type: string
+      returned: success
+    app_path:
+      description: The full path to the application folder.
+      type: string
+      returned: success
+"""
+
+import copy
```
staticdev, line 100: Imports should come right after #!/usr/bin/python, in this case is not a good idea to declare variables before imports.

staticdev (changes requested):
Sorry also for taking long and welcome back! From my side the functional side looks good. I found a couple of issues with the imports, some are super minor/cosmetic but anyway they should be on top level.

aalaesar:
> I'm not convinced only by naming for "shipped" apps, I'd personally rename it to "pre-installed", "default", "enabled_by_default' or similar. Here in documentation https://docs.nextcloud.com/server/latest/admin_manual/apps_management.html#apps-management they call them: `default enabled` or `enabled by default`.
> 
> But it's up to you, so I leave approve for your code :)

Hello there ! 😃 
Thank you for your feedback .
Regarding your 2 points:
I'm simply reusing the terminology used by the occ cli to list or filter the apps that comes preinstalled with the server, but I'm not saying it's the best word.
Regarding pre installed apps I think they are not all pre-enabled, so its hard to tell which one was enabled at any given time, as the list of enabled apps is live, and there is no filter of my knowledge to list de pre enabled apps, I can't  generate this info.

aalaesar:
> Sorry also for taking long and welcome back! From my side the functional side looks good. I found a couple of issues with the imports, some are super minor/cosmetic but anyway they should be on top level.

Hello There ! :smiley:
Thank you for your feedback.
Regarding your comments we will have to decide between official  [ansible doc regarding module code format](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#developing-modules-documenting) or python formatting good practices :confused: 
Regards
Aal.

wiktor2200:
> I'm simply reusing the terminology used by the occ cli to list or filter the apps that comes preinstalled with the server, but I'm not saying it's the best word.
Regarding pre installed apps I think they are not all pre-enabled, so its hard to tell which one was enabled at any given time, as the list of enabled apps is live, and there is no filter of my knowledge to list de pre enabled apps, I can't generate this info.

Ok, so let's stick to one naming convention or we would to change it everywhere otherwise (we still can in the future).
And that's true, pre-enabled apps are hard to determine and I think it also has changed over time in official Nextcloud repo, so I agree with current solution. :)

staticdev:
> > Sorry also for taking long and welcome back! From my side the functional side looks good. I found a couple of issues with the imports, some are super minor/cosmetic but anyway they should be on top level.
> 
> Hello There ! 😃 Thank you for your feedback. Regarding your comments we will have to decide between official [ansible doc regarding module code format](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#developing-modules-documenting) or python formatting good practices 😕 Regards Aal.

Oh my, this documentation really surprises me. Looks like it is stopped in time. In this case, let's keep your solution. If I have some time i will create an issue on Ansible side.

staticdev (approved):

plugins/modules/app_info.py
aalaesar: I think that part was formatted by only-black

aalaesar:
closing in favor of #367

PR #271, opened 2023-05-12
Fix an issue with Ubuntu Server 23.04 where download form App-store i…

marioqxx:
…ssues an error when the app is already installed, but disabled. Also no download is then necessary.

### DIFF
--- a/roles/install_nextcloud/tasks/nc_apps.yml
+++ b/roles/install_nextcloud/tasks/nc_apps.yml
@@ -19,7 +19,9 @@
   ansible.builtin.command: php occ app:install "{{ nc_app_name }}"
   args:
     chdir: "{{ nextcloud_webroot }}"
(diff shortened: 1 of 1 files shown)


marioqxx:
I do have another pull-request in my queue and wanted to add this further pull-request. This somehow messed everything up and combined the pull-requests - whcih concern very different subjects and shall not be mixed. Also because the next pull-request proposal has significant more changes. I still struggle with git and github, but its getting better ...

wiktor2200 (approved):

wiktor2200:
Hi @marioqxx! Sorry for late response. Approved, code looks good. Looking forward for your next PR.
BTW, the easiest way to work with multiple forks and branches is to checkout from main. It could give you some conflicts but should be relatively easy to solve (automatically in most cases) :)

PR #273, opened 2023-05-20
Workaround for user_saml-application to return lower-case realm when configured through environment-variable.

marioqxx:


marioqxx:
Not sure what happened, but almost all of the errors from "ansible-lint" refer to naming-convention issues, which seem to have its origin in a change of the ansible-lint checks. Almost all errors come from something other than the patch.

staticdev:
@marioqxx regarding ansible-lint it is related to an update. Since there are a lot of errors I will issue an ignore patch for now.

aalaesar:
Hello there @marioqxx 
Thank you for your enthousiasm for contributing to the collection. I really appreciate that. 👍 

However, 🤔  I feels that this  part of code is more specific to your deployment than a generic installation process.
Thoses tasks would totally make sense in a larger playbook that is runinng the installation role, then those patch tasks.
Unles a patch has been already submitted to the SAML  application repository, I think this part may need to be removed in the near future.

regards.

marioqxx:
@staticdev: Done. Thank you.

@aalaesar: You raise a valid question, I expected this question and appreciate you taking your time on this topic. Moreover, I questioned myself whether or not to add this here. I did conclude that it might be helpful and thus went forward. Certainly, you may come to a different conclusion and I'm OK with that too.

So what is the use-scenario: You want single-sign-on and administer your users centrally via LDAP, such as through OpenLDAP or Active directory. In my use-case I run a samba domain controller and provide single-sign-on for nextcloud as well as central user-administration by the samba domain-controller. Single-sign-on is then accomplished through user_saml-module. I cannot judge how often such a setup exists, because NC has put this behind a paywall. I do respect NC's decision and I do have an interest that NC is a prosperous company as well!

The issue that this patch works around is there since at least 2017 (see the date of the case I refer in the README) and it pops up on several further places in the github.com issue-tracker for the user_saml project. The most recent entry I think is from 2022/2021 and basically it concludes that this won't be changed.

Therefore, the workaround more or less becomes a permanent solution and it cannot be expected that it is not needed anymore in the near future.

In order to make this work-around work properly, I have decided _not_ to use ansible-patch module, but instead have chosen to use blockinfile. The way it is implemented makes it resilient to changes and the script would work for any user_saml-Version back till 2017 (and maybe earlier). I think it is safe to assume that the same holds true for future versions, including changes in user_saml-app that also change SAMLController.php.

The idea then to include it in this script, but not outside in another playbook was also driven by the thought, that it might be helpful for others and furthermore, the patch requires some pre-requisites, which can be easily checked in this ansible-script but would require duplicate but consistent configuration settings when placed outside. E.g. "nextcloud-root-directory", check presence of user_saml application and the like. In summary, from implementation point of view it nicely fits in here.

I also thought, that it is an quite compact and local change to this ansible script and which does not need extra servicing. It then also adds only a single configuration parameter and furthermore, the README adds value in the way that it explains an issue and helps others, avoiding them lengthy trouble-shooting and searching or even giving up at all.

staticdev:
@marioqxx I don't know the final decision from @aalaesar about adding, code LGTM. As for docs, this explanation you just did in the last comment about use-scenario would be very good to have in the new session of the install_nextcloud's README you added.

marioqxx:
@staticdev: I've updated the README according your suggestion.

aalaesar:
Hello there !
All right @marioqxx , you've made your point, that's a pitty that this need this hack to works, and to repeat it for all the application updates. 😞

roles/install_nextcloud/tasks/main.yml
```diff
@@ -122,3 +125,26 @@
...
+  when:
+    - nextcloud_patch_user_saml_app
+    - nc_available_apps.enabled['user_saml'] is defined or nc_available_apps.disabled['user_saml'] is defined
+  tags:
+    - patch_user_saml_app
+  vars:
+    pattern: '.*\$this->session->set\(.user_saml\.samlUserData.,\s+\$_SERVER\);.*'
+    file2patch: "{{ nextcloud_webroot + 'apps/user_saml/lib/Controller/SAMLController.php' }}"
+  block:
+    - name: "[Main] - Check if SAMLController.php can be patched."
```
aalaesar, line 139: Hmm I'm missing context here.
Was the following tasks (blockinfile) not idempotent for this check needed to be made ?

marioqxx: Thanks @aalaesar! This works around a behaviour of blockinfile that is not desirable in this case. The patch is only carried out when the pattern-line can be found in SAMLController.php. Despite this being in "insertbefore" the blockinfile module irrespective of whether or not the pattern is matched inserts the block in the file. If the pattern is not found it simply adds the block at the end of the file. That is why you need to make the blockinfile execution conditional on whether the pattern can be found. This is done by reading in the file and check for pattern-match in the "when"-conditional and execute blockinfile only when the pattern is present. So hope this explains this quirky implementation. In my personal code I would add a comment to the source. Here I would add a comment explaining that: "# blockinfile adds the block at the end of the file if pattern is not found. This is undesired here.". I can add this here as well if desired.

aalaesar: Ok I understand  thank you.
so it's a kind of failsafe in case the culprit code for this behaviour is removed.
LGTM

aalaesar (approved):

staticdev (approved):

PR #276, opened 2023-05-22
Temporary ignore ansible-lint var-naming[no-role-prefix]

staticdev:
This error started with last update of ansible-lint. It triggers more than 80 errors. I am adding this ignore until we fix it.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,0 +1,3 @@
+---
+skip_list:
+  - var-naming[no-role-prefix]


aalaesar (approved):
LGTM
but do we need to fix it ? 😄 
I suppose It will come down to ansible galaxy linter that'll impact the collection stars level

staticdev:
> LGTM but do we need to fix it ? smile I suppose It will come down to ansible galaxy linter that'll impact the collection stars level

I am pretty sure Galaxy does not update ansible-lint as fast as we do. But yes we should fix, I can do that soon. For now it will avoid annoying emails in our inboxes. Could you approve again? I did a failed test to fix.

aalaesar:
> > LGTM but do we need to fix it ? smile I suppose It will come down to ansible galaxy linter that'll impact the collection stars level
> 
> I am pretty sure Galaxy does not update ansible-lint as fast as we do. But yes we should fix, I can do that soon. For now it will avoid annoying emails in our inboxes. Could you approve again? I did a failed test to fix.

Actually, It reminds me of the issues I got with pre-commit hooks: #207 that got fixed with a simple update
some tests seems to  be related to ansible lint version and maybe are disabled later.
IMHO, this lint rule seems pretty arbitrary.
I mean , enforcing naming convention of variable is pretty invasive and a very opiniatred rule (that some may want to apply)

I expect it to be disabled in a future version but for now we will deal with it with the lint ignore

aalaesar (approved):

staticdev:
> > > LGTM but do we need to fix it ? smile I suppose It will come down to ansible galaxy linter that'll impact the collection stars level
> > 
> > 
> > I am pretty sure Galaxy does not update ansible-lint as fast as we do. But yes we should fix, I can do that soon. For now it will avoid annoying emails in our inboxes. Could you approve again? I did a failed test to fix.
> 
> Actually, It reminds me of the issues I got with pre-commit hooks: #207 that got fixed with a simple update some tests seems to be related to ansible lint version and maybe are disabled later. IMHO, this lint rule seems pretty arbitrary. I mean , enforcing naming convention of variable is pretty invasive and a very opiniatred rule (that some may want to apply)
> 
> I expect it to be disabled in a future version but for now we will deal with it with the lint ignore

Hopefully yes or all roles using it will have to change all task names (bit big effort if you sum up for a not so big improvement).

PR #280, opened 2023-06-03
Fix a bug in nc_apache2.j2 template concerning intermed-CA. Remove an…

marioqxx:
…sible-lint warnings and add config-variable needed for SSO to nc_apache2.j2 template. The diff in tls_signed.yml looks a bit wired, as the change involved swapping two tasks only and adding some "when" conditionals.

### DIFF
--- a/roles/install_nextcloud/tasks/tls_signed.yml
+++ b/roles/install_nextcloud/tasks/tls_signed.yml
@@ -54,5 +54,5 @@
     src: "{{ nextcloud_tls_src_cert_key }}"
     owner: "{{ nextcloud_websrv_user }}"
     group: "{{ nextcloud_websrv_group }}"
-    mode: u=r,g=-,o=-
+    mode: 0400
     force: false
--- a/roles/install_nextcloud/tasks/tls_signed.yml
+++ b/roles/install_nextcloud/tasks/tls_signed.yml
@@ -12,13 +12,13 @@
     nextcloud_tls_chain_file: "{{ nextcloud_tls_chain_file | default(cert_path + nextcloud_instance_name + \".pem\") }}"
   when: nextcloud_tls_src_chain is defined
 
-- name: "[SIGNED TLS] - Copy certificate to the host"
+- name: "[SIGNED TLS] - Copy certificate file for apache2 to the host"
   ansible.builtin.copy:
     dest: "{{ nextcloud_tls_cert_file }}"
     src: "{{ nextcloud_tls_src_cert }}"
     owner: "{{ nextcloud_websrv_user }}"
     group: "{{ nextcloud_websrv_group }}"
-    mode: "0640"
+    mode: u=rw,g=r,o=-
     force: true
   when:
     - nextcloud_websrv not in ["nginx"]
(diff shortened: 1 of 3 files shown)


staticdev (approved):
LGTM.

wiktor2200:
Hi! These changes for TLS certs looks good, but I think we shouldn't change `mode` to different notation. We should keep it consistent across the repository. So I suggest reverting `mode` changes.

roles/install_nextcloud/tasks/tls_signed.yml
```diff
@@ -12,13 +12,13 @@
...
 
-- name: "[SIGNED TLS] - Copy certificate to the host"
+- name: "[SIGNED TLS] - Copy certificate file for apache2 to the host"
   ansible.builtin.copy:
     dest: "{{ nextcloud_tls_cert_file }}"
     src: "{{ nextcloud_tls_src_cert }}"
     owner: "{{ nextcloud_websrv_user }}"
     group: "{{ nextcloud_websrv_group }}"
-    mode: "0640"
+    mode: u=rw,g=r,o=-
```
wiktor2200, line 21:
```suggestion
    mode: "0640"
```

roles/install_nextcloud/tasks/tls_signed.yml
```diff
@@ -29,28 +29,30 @@
     dest: "{{ nextcloud_tls_cert_file }}"
     owner: "{{ nextcloud_websrv_user }}"
     group: "{{ nextcloud_websrv_group }}"
-    mode: "0640"
+    mode: u=rw,g=r,o=-
```
wiktor2200, line 32:
```suggestion
    mode: "0640"
```

roles/install_nextcloud/tasks/tls_signed.yml
```diff
@@ -29,28 +29,30 @@
...
+- name: "[SIGNED TLS] - Copy certificate chain file for apache2 to the host"
   ansible.builtin.copy:
-    dest: "{{ nextcloud_tls_cert_key_file }}"
-    src: "{{ nextcloud_tls_src_cert_key }}"
+    dest: "{{ nextcloud_tls_chain_file }}"
+    src: "{{ nextcloud_tls_src_chain }}"
     owner: "{{ nextcloud_websrv_user }}"
     group: "{{ nextcloud_websrv_group }}"
-    mode: "0400"
+    mode: u=rw,g=r,o=-
```
wiktor2200, line 45: Have you changed it from 0400 to 0640 on purpose here?

roles/install_nextcloud/tasks/tls_signed.yml
```diff
@@ -29,28 +29,30 @@
...
+- name: "[SIGNED TLS] - Key is copied to the host"
   ansible.builtin.copy:
-    dest: "{{ nextcloud_tls_chain_file }}"
-    src: "{{ nextcloud_tls_src_chain }}"
+    dest: "{{ nextcloud_tls_cert_key_file }}"
+    src: "{{ nextcloud_tls_src_cert_key }}"
     owner: "{{ nextcloud_websrv_user }}"
     group: "{{ nextcloud_websrv_group }}"
-    mode: "0640"
+    mode: u=r,g=-,o=-
```
wiktor2200, line 57: Have you changed it from 0400 to 0640 on purpose here?

wiktor2200 (changes requested):

staticdev:
> Hi! These changes for TLS certs looks good, but I think we shouldn't change `mode` to different notation. We should keep it consistent across the repository. So I suggest reverting `mode` changes.

I also prefer that the whole code has consistent modes, good point.

roles/install_nextcloud/tasks/tls_signed.yml
marioqxx: Ouch, it should be "0400" for the key! You are right! Since I've signed-off your proposal for the mode above, should I change this or is it better you propose the change?

roles/install_nextcloud/tasks/tls_signed.yml
marioqxx: Actuallly the change is from 0640 to 0400. Here 0640 is appropriate, this is a certificate, not a key. So seems I mixed this with the former permissions up ... . As before, is it better your propose the change to the mode or should I do this change?

wiktor2200 (changes requested):

staticdev:
@wiktor2200 @marioqxx after this one we can release 2.1.0 ;)

wiktor2200 (approved):

wiktor2200:
@marioqxx Approved and merged :)

wiktor2200:
@marioqxx Approved and merged :)
@staticdev we can proceed with release I suppose ;)

PR #285, opened 2023-06-19
Feature default NC27

staticdev:


molecule/default/latest-converge.yml
```diff
@@ -15,7 +15,7 @@
     nextcloud_apps:
       calendar:
       contacts:
-        source: "https://github.com/nextcloud-releases/contacts/releases/download/v5.2.0/contacts-v5.2.0.tar.gz"
+        source: "https://github.com/nextcloud-releases/contacts/releases/download/v5.3.1/contacts-v5.3.1.tar.gz"
```
wiktor2200, line 18: Version 5.3.1 is broken in releases. (it was released with capital V5.3.1)
But it's already fixed in 5.3.2
```suggestion
        source: "https://github.com/nextcloud-releases/contacts/releases/download/v5.3.2/contacts-v5.3.2.tar.gz"
```

wiktor2200: @staticdev I can see that this change disappeared with last commit/push.

staticdev: I think we discovered that at the same time ;)

staticdev:
@wiktor2200 I am having a strange behavior.. locally it is working good the playbook, but on GA not. Could you try to run  `molecule test` locally?

wiktor2200:
Hi @staticdev! The same on my side. I confirm that it works locally, but on GA I'm having errors. I've also noticed, that errors also showing up in other steps (such as linter).
First attempt failed: https://github.com/nextcloud/ansible-collection-nextcloud-admin/actions/runs/5363160101/attempts/1
But then after re-play it started working: https://github.com/nextcloud/ansible-collection-nextcloud-admin/actions/runs/5363160101

staticdev:
@wiktor2200 then, I also suggest to merge and watch. I will start working now on Debian 12 upgrade ;)

wiktor2200 (approved):
Hi @staticdev! I totally agree, merging this one will solve at least problems with tests and molecule. 
GA related problems may be fixed automatically by GA contributors/maintainers and maybe we will get affected :)

PR #288, opened 2023-06-26
Support Debian 12

staticdev:
This one was easier than expected, just works =)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        distro: [debian11, debian10, ubuntu2204, ubuntu2004]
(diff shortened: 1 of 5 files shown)


wiktor2200 (approved):
Hi @staticdev! I'm glad to hear good news here :)
LGTM

staticdev:
Thank @wiktor2200, I am just very annoyed about molecule failing all the time. For my other repos I don't have this problem but I use molecule with podman instead of docker. I remember podman didn't work for install nextcloud on first try, maybe I can try to troubleshoot that.

wiktor2200:
Yeah, it's quite annoying. What is more it's sending lots of emails about failed pipelines. Unfortunately I won't have time this week to also take a closer look on that, so I'm keeping my fingers crossed for your trial :)

PR #292, opened 2023-07-07
Add possibility to use a downloaded nextcloud archive and copy on rem…

marioqxx:
…ote host. This is useful for setting up config or when installing on multiple machines, because it avoids multiple downloads.

### DIFF
--- a/roles/install_nextcloud/README.md
+++ b/roles/install_nextcloud/README.md
@@ -556,7 +556,7 @@ nextcloud_apps:
       parameter2: another_value
   app_name_1: # Copy from control-host.
     source: "files/some_archive.zip"
-  app_name_1: # Download from nextcloud official app-store
+  app_name_3: # Download from nextcloud official app-store
     conf:
       parameter1: ldap:\/\/ldapsrv
       parameter2: another_value
--- a/roles/install_nextcloud/README.md
+++ b/roles/install_nextcloud/README.md
@@ -116,7 +116,7 @@ Choose between the 2 archive formats available in the repository.
 ```
 
 _If you don't like rules..._
-Specify directly a full URL to the archive. The role will skip the url generation and download the archive. **Requires nextcloud_version_major to be set along**.
+Specify directly a path or full URL to the archive. The role will skip the url generation and copy from the control-host or download the archive. **Requires nextcloud_version_major to be set along**.
 
 #### Examples:
 - Download your own archive:
--- a/roles/install_nextcloud/README.md
+++ b/roles/install_nextcloud/README.md
@@ -112,19 +112,24 @@ nextcloud_archive_format: "zip" # zip | tar.bz2
 Choose between the 2 archive formats available in the repository.
 
 ```yaml
-# nextcloud_full_url:
+# nextcloud_full_src:
 ```
 
 _If you don't like rules..._
 Specify directly a path or full URL to the archive. The role will skip the url generation and copy from the control-host or download the archive. **Requires nextcloud_version_major to be set along**.
 
 #### Examples:
-- Download your own archive:
+- Download or Copy your own archive:
   (_you **must** specify the nextcloud major version along_)
 
 ```yaml
-nextcloud_full_url: https://download.nextcloud.com/server/releases/nextcloud-25.0.0.zip
-nextcloud_version_major: 42
+# Download via URL
+nextcloud_full_src: https://download.nextcloud.com/server/releases/nextcloud-25.0.0.zip
+nextcloud_version_major: 25
+# Copy from control host
+nextcloud_full_src: "files/nextcloud-27.0.0.zip"
+nexcloud_version_major: 27
+
 ```
 
 -   Choose the latest release (default):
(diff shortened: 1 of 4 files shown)


marioqxx:
On my compiter, ansible-lint does not throw any error. Also, I have tested all variants and it worked. Changes to code are very view only, thus should not show any issue.

aalaesar:
> On my computer, ansible-lint does not throw any error. Also, I have tested all variants and it worked. Changes to code are very view only, thus should not show any issue.

Hello there 

there is an issue with the github action docker setup. not sure if it comes from github  or our GA config

marioqxx:
Thanks for clarifying @aalaesar. Good to know that this is not an issue on my end.

roles/install_nextcloud/README.md
```diff
@@ -112,19 +112,24 @@ nextcloud_archive_format: "zip" # zip | tar.bz2
...
 
 ```yaml
-nextcloud_full_url: https://download.nextcloud.com/server/releases/nextcloud-25.0.0.zip
-nextcloud_version_major: 42
+# Download via URL
+nextcloud_full_src: https://download.nextcloud.com/server/releases/nextcloud-25.0.0.zip
+nextcloud_version_major: 25
+# Copy from control host
+nextcloud_full_src: "files/nextcloud-27.0.0.zip"
+nexcloud_version_major: 27
```
wiktor2200, line 131: No need to duplicate the code.
`nexcloud_version_major` can be moved above these 2 variables.

roles/install_nextcloud/README.md
```diff
@@ -531,23 +536,26 @@ Since **v1.3.0**, it is possible to download, install and enable nextcloud appli
 
 The application (app) to install have to be declared in the `nextcloud_apps` dictionary in a "key:value" pair.
 -   The app name is the key
--   The download link, is the value. If left blank, the app is tried to be downloaded and installed from the nextcloud official app-store. 
+-   The file path or download link, is the value. If left blank, the app is tried to be downloaded and installed from the nextcloud official app-store. If a file path is provided, the app is copied from that path of the control-host.
```
wiktor2200, line 539: I would mention `download links` first as default installation method.
```suggestion
-   The download link or file path is the value. If left blank, the app is tried to be downloaded and installed from the nextcloud official app-store. If a file path is provided, the app is copied from that path of the control-host.
```

roles/install_nextcloud/README.md
```diff
@@ -531,23 +536,26 @@ Since **v1.3.0**, it is possible to download, install and enable nextcloud appli
...
 
 ```yaml
 nextcloud_apps:
-  app_name_1:
+  app_name_1: # Download from specified URL.
     source: "http://download_link.com/some_archive.zip"
     conf:
       parameter1: ldap:\/\/ldapsrv
       parameter2: another_value
+  app_name_1: # Copy from control-host.
```
wiktor2200, line 557:
```suggestion
  app_name_2: # Copy from control-host.
```

roles/install_nextcloud/README.md
```diff
@@ -531,23 +536,26 @@ Since **v1.3.0**, it is possible to download, install and enable nextcloud appli
...
+    source: "files/some_archive.zip"
   app_name_1: # Download from nextcloud official app-store
```
wiktor2200, line 559:
```suggestion
  app_name_3: # Download from nextcloud official app-store
```

roles/install_nextcloud/README.md
wiktor2200: I've left this comment for consistency with variables above.

roles/install_nextcloud/tasks/nc_apps.yml
```diff
@@ -25,11 +25,11 @@
   register: output
   changed_when: "output.rc == 0"
 
-- name: "[App] - Download Archive in apps folder from \"{{ nc_app_source }}\""
+- name: "[App] - Copy or Download Archive to apps folder from \"{{ nc_app_source }}\""
```
wiktor2200, line 28:
```suggestion
- name: "[App] - Download or copy Archive to apps folder from \"{{ nc_app_source }}\""
```

roles/install_nextcloud/defaults/main.yml
```diff
@@ -15,7 +15,7 @@ nextcloud_get_latest: true  # mandatory # specify if the latest archive should b
 # nextcloud_version_special: "" #  For prereleases: "RCn/beta" or for daily "YYYY-MM-DD"
 nextcloud_repository: "https://download.nextcloud.com/server"  # Domain URL where to download Nextcloud.
 nextcloud_archive_format: "zip"  # zip/tar.bz2
-# nextcloud_full_url: "https://download.nextcloud.com/server/releases/nextcloud-25.0.0.zip"  # specify directly a full URL to the archive
+# nextcloud_full_src: "https://download.nextcloud.com/server/releases/nextcloud-25.0.0.zip"  # specify directly a path on the control host or full URL to the archive
```
wiktor2200, line 18:
```suggestion
# nextcloud_full_src: "https://download.nextcloud.com/server/releases/nextcloud-25.0.0.zip"  # specify directly a full URL to the archive or  a path on the control host
```

wiktor2200 (changes requested):
Hi! @marioqxx sorry for late response, but I've reviewed your code.
I really like this feature and it makes lots of sense for multiple installations at the same time.
I've left some comments - mostly editorial and cosmetic.

roles/install_nextcloud/README.md
marioqxx: This is correct. My intention was to indicate to the reader, that `nextcloud_version_major` must always be set, even though ansible could derive the major version from the filename. Your comment evidences, that this did not work out. Moving up would not work, because on the one hand side, there is version 25 and on the other 27. Have you got any suggestion how to improve the description?

wiktor2200: Got it! It makes sense in this case if it's just to emphasize. And as it's only the README we can treat it as example.

wiktor2200 (approved):
LGTM

PR #304, opened 2023-08-26
Change role description

wiktor2200:
I would propose description change for Ansible Collection. In my opinion having it mentioned that it's Official under Nextcloud organization may be beneficial.

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -21,7 +21,7 @@ authors:
 
 ### OPTIONAL but strongly recommended
 # A short summary description of the collection
-description: A collection of ansible roles and modules to administrate nextcloud instances
+description: Official Nextcloud Ansible collection of roles and modules to deploy and manage Nextcloud instances
 
 # The path to the license file for the collection. This path is relative to the root of the collection. This key is
 # mutually exclusive with 'license'


aalaesar (approved):
LGTM

staticdev:
You are toont fast for me these days sorry. Super busy here and also struggling with losing files from my nextcloud instalation (trying to understand why)

wiktor2200:
Hi @staticdev! No problem at all, @aalaesar has already approved :)

Oh.. your problem sounds serious. I had similar problem in the past (for NC 20 AFAIR) and the problem was connected with one desktop client's old version. It had some problems with syncing, so when I added one file in one client, the other removed that.
But files were in "Deleted" folder, so I was able to restore. Start with checking your retention settings and client's logs (if you have such possibility).

AndyScherzinger:
I am also a bit to late than... ;)

I would have suggested to drop the "Official " keyword, so people won't misinterpret it because Nextcloud (as a company) also offers support to paying customer. And the word "official" might suggest it is also part of what is supported under the payed subscription (spoiler: this isn't). Also currently no harm in having it in, I am aware you mean official in a different way - like under the Nextcloud umbrella/org.

aalaesar:
Hello @AndyScherzinger ,
That's a fair point.
 Fortunately, the change has not been released yet so we can still find a not ambiguous  way to say that this is  a nextcloud  community  project

wiktor2200:
Hi! Yes I meant `official in a different way - like under the Nextcloud umbrella/org`. So maybe to differentiate that it would be good to have: `Official community` to make sure that's still one of the ways, supported by community and not paid/enterprise solution. :thinking:

wiktor2200:
Hi! Yes I meant `official in a different way - like under the Nextcloud umbrella/org`. So maybe to differentiate that it would be good to have: `Official community` to make sure that's still one of the ways, supported by community and not paid/enterprise solution.

PR #318, opened 2023-11-11
Fix GitubActions Docker issue

wiktor2200:


wiktor2200:
Hi! @staticdev @aalaesar I've tried to fix a problem with docker molecule test but I got run out of idea. Would you be able to take a look and see? Maybe you will have some other solutions for this problem?

@geerlingguy Sorry for bothering you, but maybe you have got any idea what could have gone wrong here? Have you ever seen such error in yours Ansible images?

aalaesar:
Hello @wiktor2200 thank you for taking some time to fix the CI.
I've also been trying to fix it on some other branch but with no success. :disappointed: 
Most of the time I suppose the issue is in our code as the ansible image used is popular and I couldn't find someone with a similar issue. 
Let see if Jeff Geerling can help us :wink: 

Regards

geerlingguy:
It looks like the error is:

```
Error creating container: 500 Server Error for http+docker://localhost/v1.43/containers/create?name=instance: Internal Server Error (\"symlink /proc/mounts /var/lib/docker/fuse-overlayfs/4441cd54c476cdd29d6f1ded1e93781e3c3929ca7407bbc645bd90b92c4c22e2-init/merged/etc/mtab: file exists\")
```

I've seen similar file mount issues in GitHub Actions sometimes, but haven't in the past few months. Is this only with debian12?

wiktor2200:
Hello Jeff! thanks a lot for involvement, I really appreciate that :)

When we were searching for this issue, there is not many issues found, that's why I asked.
**It occurs randomly in all of our Molecule tests scenarios (both Debian 11,12 and Ubuntu20.04, 22.04)**, we define scenarios this way:
https://github.com/nextcloud/ansible-collection-nextcloud-admin/blob/20ab659c9d5eeaef1d091d4571bb17623e47edb8/.github/workflows/tests.yml#L20-L23

Then running it with: https://github.com/nextcloud/ansible-collection-nextcloud-admin/blob/20ab659c9d5eeaef1d091d4571bb17623e47edb8/.github/workflows/tests.yml#L51

Molecule itself it defined here:
https://github.com/nextcloud/ansible-collection-nextcloud-admin/blob/20ab659c9d5eeaef1d091d4571bb17623e47edb8/molecule/default/molecule.yml#L7-L15

And as it's matrix when once fails, rest are cancelled. In this PR I've tried to clean docker cache (inspired with your old blog post: https://www.jeffgeerling.com/blog/2018/testing-your-ansible-roles-molecule) and then `molecule reset` when `docker system prune` didn't help.

aalaesar:
Hello there
@wiktor2200 found [this](https://discuss.linuxcontainers.org/t/docker-container-fails-sometimes-with-symlink-proc-mounts-file-exists-with-fuse-overlayfs/17449) subject on lxc forum that is looping much like our issue.
Is there a way to check if our github actions are running on top of of LXC?

staticdev:
hello all, I have been super busy with some other ansible issues, construction (like @geerlingguy =p) and don't really understand why this issue is happening. most my roles are tested against the same images and I don't have such an error. I would say to try using podman instead of docker since I mostly replaced docker for podman now. it is an alternative solution

.github/workflows/tests.yml
```diff
@@ -44,9 +44,9 @@ jobs:
 
       - name: Install test dependencies
         run: |
-          python3 -m pip install --constraint=.github/workflows/constraints.txt ansible 'molecule-plugins[docker]' docker netaddr
+          python3 -m pip install --constraint=.github/workflows/constraints.txt ansible 'molecule-plugins[podman]' podman netaddr
```
staticdev: adding `podman` is not necessary

.github/workflows/lint.yml
```diff
@@ -16,8 +16,7 @@ jobs:
           fetch-depth: 0
 
       - name: Run ansible-lint
-        # replace `main` with any valid ref, or tags like `v6`
-        uses: ansible/ansible-lint-action@v6
+        uses: ansible/ansible-lint@v6.22.0
```
staticdev, line 19: This is making some new complains about linting.. at least I see the logs from latest execution showing 2 errors. I would keep it separate from this PR.

.github/workflows/tests.yml
```diff
@@ -28,6 +28,10 @@ jobs:
         with:
           path: "nextcloud.ansible-collection-nextcloud-admin"
 
+      - name: Clean docker cache
```
staticdev, line 31: Those lines also do not make sense when you are testing podman.

staticdev (changes requested):

staticdev:
@wiktor2200 Thanks for trying it out. I saw some potential issues with current state of the PR on comments.

aalaesar:
Hello there.
I noticed that we are not running  on this issue anymore now..... somehow the issue disapeared...
I'll keep the Pr in draft for now until we are confident the issue is gone for good.
Regards

aalaesar:
initial issue is gone now and CI has been fixed to work now. closing

PR #336, opened 2024-01-13
Fix a bug in nc_apps when conf is present but no source. This is not …

marioqxx:
…needed any more, because in this case the install is from the official App-store. Additionally fix an slight issue with Debian Bookworm for some custom install-variants. This is related to https://github.com/nextcloud/ansible-collection-nextcloud-admin/issues/335

### DIFF
--- a/roles/install_nextcloud/defaults/main.yml
+++ b/roles/install_nextcloud/defaults/main.yml
@@ -32,6 +32,7 @@ php_pkg_spe:
   - "php{{ php_ver }}-mbstring"
   - "php{{ php_ver }}-xml"
   - "php{{ php_ver }}-zip"
+  - "php{{ php_ver }}-apcu"
 php_socket: "/run/php/php{{ php_ver }}-fpm.sock"
 php_memory_limit: 512M
 
--- a/roles/install_nextcloud/defaults/main.yml
+++ b/roles/install_nextcloud/defaults/main.yml
@@ -32,7 +32,6 @@ php_pkg_spe:
   - "php{{ php_ver }}-mbstring"
   - "php{{ php_ver }}-xml"
   - "php{{ php_ver }}-zip"
-  - "php{{ php_ver }}-apcu"
 php_socket: "/run/php/php{{ php_ver }}-fpm.sock"
 php_memory_limit: 512M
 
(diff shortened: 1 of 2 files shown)


roles/install_nextcloud/tasks/nc_apps.yml
```diff
@@ -1,17 +1,10 @@
 ---
-- name: nc_apps | Parse the item values
+- name: "nc_apps | Parse the item values"
```
staticdev, line 2: The quotes here are not necessary since you are not templating anything.

staticdev (changes requested):

roles/install_nextcloud/defaults/main.yml
```diff
@@ -32,6 +32,7 @@ php_pkg_spe:
   - "php{{ php_ver }}-mbstring"
   - "php{{ php_ver }}-xml"
   - "php{{ php_ver }}-zip"
+  - "php{{ php_ver }}-apcu"
```
staticdev, line 35: This is already installed in [php_install.yml](https://github.com/nextcloud/ansible-collection-nextcloud-admin/blob/8f5d564bcedc603205d4346db21388a390689419/roles/install_nextcloud/tasks/php_install.yml#L42)

staticdev (changes requested):

roles/install_nextcloud/defaults/main.yml
marioqxx: OK, interesting. I had some issue with Debian/Ubuntu fresh-installs and only some packages selected for pre-installation. Would you want me to remove this, it otherwise shouldn't hurt.

roles/install_nextcloud/tasks/nc_apps.yml
staticdev: Actually you are making more inconsistent. We should **consistently** use quotes only when templating.

roles/install_nextcloud/defaults/main.yml
staticdev: Yes please.

marioqxx: Done - at least for me it looks like this. In case not, this is attributable to my insufficient git-knowledge.

roles/install_nextcloud/tasks/nc_apps.yml
marioqxx: Got it, I haven't realized this. Have changed it.

staticdev (approved):

staticdev:
Thanks @marioqxx !

PR #356, opened 2024-06-02
A new role, which downloads the archive to the control host. This all…

marioqxx:
…ows then to use the install-feature, that downloaded archives are transferred to the remote-host instead of downloading. This is the intial attempt. Therefore, please review and provide your comments.

### DIFF
--- a/roles/download/tasks/nc_apps.yml
+++ b/roles/download/tasks/nc_apps.yml
@@ -1,0 +1,38 @@
+---
+- name: nc_apps | Parse the item values
+  ansible.builtin.set_fact:
+    nc_app_name: "{{ item.value if item.value!=None else item.key }}"
+
+- name: "{{ 'nc_apps | Fetch web-page for app \"' ~ item.key ~'\" from online-apps store (1/2).' }}"
+  ansible.builtin.get_url:
+    url: "{{ nextcloud_app_repository }}/{{ nc_app_name }}"
+    dest: /tmp/app.html
+    mode: u=rw,g=r,o=r
+    force: true
+  changed_when: false
+  delegate_to: localhost
+  become: false
+
+# Retrieve the app-package archive from the web-page content.
+- name: "{{ 'nc_apps | Fetch for app \"' ~ item.key ~ '\" package download-link from web-page and download from online-apps store (2/2).' }}"
+  ansible.builtin.get_url:
+    url: "{{ dl_link }}"
+    dest: "{{ nextcloud_download_dir|default('.') ~ '/' ~ dl_link|basename }}"
+    mode: u=rw,g=r,o=r
+  delegate_to: localhost
+  become: false
+  vars:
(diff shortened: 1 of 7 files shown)


aalaesar:
hello there @marioqxx .
Thank you for still contributing to the collection.
I have reviewed the role and it's already well done,documented and integrated with the same variables as install_nextcloud.

There is definitively integration to do with it in the main role already.

During reading It came to my mind some questions:
- could it be splitted into 2 roles (one for download server, the other one for apps) ?
- can we call it from install_nextcloud to avoid dual computation of the download links ?
- can we improve the version_major and version_full vars to merge then into one ? (using version filter instead of  using tasks definitions )
I'm not asking you to do anything regarding that, just opening discussions. 😄

regards. Aal

aalaesar:
also please rebase the branche from main: the CI has been changed

roles/download/tasks/nc_apps.yml
```diff
@@ -0,0 +1,38 @@
+---
+- name: nc_apps | Parse the item values
+  ansible.builtin.set_fact:
+    nc_app_name: "{{ item.value if item.value != None else item.key }}"
```
aalaesar, line 4: hello there.
you don't need to use set_facts for one time use variables. You can just declare it as ad-hoc var in the task below.
set_fact have a high level of precedence in the var hierarchy of ansible. furthermore, having the "fact" being changed at every iteration of the loop could create issues.
```suggestion
# add this in the task below instead
  vars:
    nc_app_name: "{{ item.value if item.value != None else item.key }}"
```

aalaesar (changes requested):

marioqxx:
@aalaesar,

thanks for looking into this! I have some issue which I need to resolve in order make my development environment work again, so please be a bit patient. I have accepted your proposed change. Good to know, this is also included in the nextcloud_install/nc_apps.yml I think. I will test your suggestion, because sometimes I have experienced issues with ansible, packing such expressions in a "var"-statement. But this again goes back to my insufficient knowledge of ansible.

To your question:
**"could it be splitted into 2 roles (one for download server, the other one for apps) ?"**
Of course. The way I have handled this is by means of the two variables:
`nextcloud_download_server: true`
`nextcloud_download_apps: true`
So dependent on their setting you can download either server or app or both.

**can we call it from install_nextcloud to avoid dual computation of the download links ?**
I don't think this would work. I had to create the computation of the download-link completely new and could not use anything from install_nextcloud. The way it is done is quite comprehensive, it includes HTML-parsing of the App-Shop WEB-page.

**can we improve the version_major and version_full vars to merge then into one ? (using version filter instead of using tasks definitions )**
You are right. I went the simplified way down the road, with an alignment to install_nextcloud. I'll have a look at it and see how this could be done better.

Once I have re-established my dev-setup, I'll potentially ask for your help regarding adding "molecule test". This is definitely outside my competence.

PR #378, opened 2025-01-17
Add func to module_utils/occ.py to convert string to list with two or much words in parameters

j0ker-222:
fix problem when run occ command with two or much words in parameter, for example: 
```
nextcloud.admin.run_occ:
  command: group:add -- "Some group name"
  nextcloud_path: "/var/www/nextcloud"
```
or
```
nextcloud.admin.run_occ:
  command: group:add -- 'Some another group name'
  nextcloud_path: "/var/www/nextcloud"
```
parameter "Some group name" reading like each different option - ['"Some', 'group', 'name"'] and not like ['"Some group name"']


j0ker-222:
sorry, got some logic mistakes

PR #379, opened 2025-01-17
Add func to module_utils/occ.py to convert string to list with two or much words in parameters

j0ker-222:
fix problem when run occ command with two or much words in parameter, for example:
```
nextcloud.admin.run_occ:
  command: group:add -- "Some group name"
  nextcloud_path: "/var/www/nextcloud"

```
or
```
nextcloud.admin.run_occ:
  command: group:add -- 'Some another group name'
  nextcloud_path: "/var/www/nextcloud"

```
parameter "Some group name" reading like each different option - ['"Some', 'group', 'name"'] and not like ['"Some group name"']
P.S.: sorry, in previous commit got logic and name mistakes

### DIFF
--- a/plugins/module_utils/occ.py
+++ b/plugins/module_utils/occ.py
@@ -24,7 +24,16 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import os
+from shlex import shlex
 
+def convert_string(command: str) -> list:
+    command_list = shlex(command, posix=True)
+    command_list.whitespace_split = True
+    if command_list.find("'") > 0:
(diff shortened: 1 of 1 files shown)


aalaesar (commented):
Hell @j0ker-222 thank you for pointing out this issue and proposing your fix.
Is there any special constraint emerging from using shlex ?
like a minimal python version etc ?

j0ker-222:
sorry, but can't say for sure, but if i correctly remember - minimal version need python 3.8, because they add some changes in function 'whitespace_split' in 3.8 vers
I test it only on python 3.9 and higher, all worked good

aalaesar (approved):
LGTM

PR #388, opened 2025-03-03
Adds support for occ commands with parameters

sebastian-schmid:
I had Issues using the run_occ tasks with commands using parameters like `command: theming:config name "myNextcloud"`.

I tried to simplify the code to make it work with all occ commands.

### DIFF
--- a/plugins/module_utils/occ.py
+++ b/plugins/module_utils/occ.py
@@ -24,16 +24,8 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import os
-from shlex import shlex
+import shlex
 
-def convert_string(command: str) -> list:
-    command_lex = shlex(command, posix=True)
-    command_lex.whitespace_split = True
-    if command.find("'") > 0:
-        command_lex.quotes = "'"
-    elif command.find('"') > 0:
-        command_lex.quotes = '"'
-    return list(command_lex)
 
 def run_occ(
     module,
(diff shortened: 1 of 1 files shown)


aalaesar:
Hello @sebastian-schmid  thank you for proposing your PR.
There are quite few changes in your PR that seem not related to "parameters issue"
Could you describe the issue you encountered  ?
Thank you.

sebastian-schmid:
Hello @aalaesar 

I had the Issue that I could not use e.g. `command: theming:config name "myNextcloud"`. 

I also removed the maintanence stuff because there was an error like "the call does only await 3 return values".

Looks like here https://github.com/nextcloud/ansible-collection-nextcloud-admin/blob/a0c0f9fd9dd48e0a8f4d99a80a5b176dbb5cf234/plugins/modules/run_occ.py#L128 is no `maintenance` defined, maybe the better way is to add it there and not to remove it?

aalaesar:
> Hello @aalaesar
> 
> I had the Issue that I could not use e.g. `command: theming:config name "myNextcloud"`.
> 
> I also removed the maintanence stuff because there was an error like "the call does only await 3 return values".
> 
> Looks like here
> 
> https://github.com/nextcloud/ansible-collection-nextcloud-admin/blob/a0c0f9fd9dd48e0a8f4d99a80a5b176dbb5cf234/plugins/modules/run_occ.py#L128
> is no `maintenance` defined, maybe the better way is to add it there and not to remove it?

That's weird.
If you can reproduce with logs , those are welcome.

The maintenance mode affect a lot how the occ cli answer and the commands available. I anticipated that returning this value would help for other future modules that could be impacted by the maintenance mode , even if the basic run_occ don't leverage it.
So more details and logs about your issue would help me review the PR 😃 

Regards

aalaesar:
Also I will take some times to fix the CI stuff that is broken

aalaesar:
@sebastian-schmid 
Run_occ have been greatly improved on parameters  parsing.
Maybe your initial  issue is also resolved with the latest commit in main ?
Could you check please ?

Thanks.

Aal

PR #426, opened 2025-10-10
add debian 13 support

aalaesar:


David-Else:
I think this also needs addressing: https://github.com/nextcloud/ansible-collection-nextcloud-admin/issues/422

PR #452, opened 2026-02-27
Add a nextcloud.admin.upgrade role to the collection

fzadikian:
Good evening. For the last month, I've been working on an upgrade role to use on some of the Nextcloud instances at the [university I work at](https://udelar.edu.uy/). Although it might not suit all use-cases (see caveats in [readme file](https://github.com/cielito/ansible-collection-nextcloud-admin/blob/upgrade-role-upstream/roles/upgrade/README.md)), it seems useful enough to propose upstream.

I can't promise to add more features in the short-term (though I can make amendments in order to hopefully merge), but I hope someone gets inspired and builds on top of this.

All pre-commit tests specified in [CONTRIBUTING.md](https://github.com/nextcloud/ansible-collection-nextcloud-admin/blob/main/CONTRIBUTING.md) passed. The role has been tested from Nextcloud 20 to 32, including automatic PHP and PostgreSQL upgrades. Our internal PR (with full commit history) is [here](https://github.com/cielito/ansible-collection-nextcloud-admin/pull/2).

Any comments/suggestions are greatly appreciated (especially regarding pre/post upgrade tasks and default variables, which may be biased in favor of our own instances). Thank you!


aalaesar:
Hello there @fzadikian 
First thank you very much for your contribution and maintaining the Request without feed back from me.

I'm sorry taking so long to come back to you. I find your role well written, and autonomous. great Job !

My issue is more about how my perspective have changed over the decade: it doesn't match anymore with the "one role do it all" I used 10 years ago.

This made me hold my review several times already but i think it's not fair you.
To go forward, I propose to validate the role as it is now.
Then I would propose later a re-work of both install and upgrade roles to modernize the collection's structure. with a better modular code structure that will enable less maintenance for future nextcloud upgrades & installations.

Regards.

Aal

fzadikian:
Hello @aalaesar,

Great! I agree with your proposal. One detail is that, as you can see, the role needs to be updated for each version (although users can set the path manually as well). Are you willing to do that on main, or how should we do it?

One way to minimize change would be to add the `maintenance:repair --include-expensive` post-upgrade task to all future versions, instead of re-adding it for each major Nextcloud version. What do you think?

aalaesar:
> Great! I agree with your proposal.

Nice :+1: 

>  One detail is that, as you can see, the role needs to be updated for each version (although users can set the path manually as well). Are you willing to do that on main, or how should we do it?

I can't promise anything. I'm working on this project on my free time (which has became rare over the years)
For future maintenance feel free to submit any update you want.
The best would be that the collection releases are not constrained by nextcloud's release schedules. 

> One way to minimize change would be to add the `maintenance:repair --include-expensive` post-upgrade task to all future versions, instead of re-adding it for each major Nextcloud version. What do you think?

I'm not sure. I didn't found much doc about what it does. this may depend on external application installed etc. This can be discussed in a new discussion.

fzadikian:
@aalaesar Alright, if you agree we can keep it this way and eventually further improve it in a separate discussion. Regarding tests, I can see some from install_nextcloud have failed. Should I write tests for the upgrade role?

fzadikian:
Hello! Yesterday I refactored the role to work with Nextcloud's major version, instead of the full version. This way, now there is no need to update the default variables for each patch release. The version checks and xml parsing were also improved. I ended up adding the `maintenance:repair --include-expensive` task to every step, since I was adding it to each new version regardless. I've tested the current code on stage and will use it to update an instance we've got on production soon. What would be the next steps to get this role merged to main? Thank you!

README.md
```diff
@@ -48,6 +48,7 @@ nextcloud.admin.group | Manage Nextcloud groups.
 
 Name | Description
 --- | ---
+nextcloud.admin.upgrade (**alpha**) | Upgrade a Nextcloud instance
```
aalaesar: Do you feel confident enough to mark it as ready ? :)

aalaesar (dismissed):

README.md
fzadikian: I do! It goes without saying that any issues regarding this role can be assigned to me.

aalaesar (approved):
Very well.
@fzadikian thank you again for your great contribution! 
(Thank you for your patience with me on merging it)

aalaesar:
Hello there
Molecule tests won't pass, they need an upgrade. (And that's alsolnot the scope of this PR) so merging now.
I'll make the 2.4.0 release once they are fixed globally.
Regards.

PR #464, opened 2026-07-14
PHP Config as modules, fpm support in apache, option for custom php config

tty-13:
Hello and thanks for the work on this collection, it's a great time-saver.

I saw the php config in the install role was directly written on the main php.ini file and thought it would be better to add all this config as modules, to make it clearer what has been changed by who.

I implemented a few tasks to bring this feature : 

 - PHP options are made available as modules in php/.../mods_available directory.
   There's a general php module and a cli-specific one.
   The normal config is loaded in `{ php_dir }/fpm/conf.d directory*` with priority 50
   The cli config is loaded in `{ php_dir }/cli/conf.d` with priority 50
- Two additional configs are loaded in the same directories with priority 55 : those are for the user-provided overrides (two new vars)
- 4 handlers were added before **Reload php-fpm**  to load all those configs if they exist.

Also, I saw fpm was only half-supported in the apache task, I added the corresponding apache modules and conf to be loaded accordingly, I guess that was missing since php-fpm gets installed instead of just php.
( a2enmod proxy_fcgi and setenvif, and a2enconf php{php_ver}-fpm )

_*(without fpm support, the php module could also be loaded in apache2, didn't implement that though, since we would first have to add some options to let the user choose between php and php-fpm)_

Would you accept this contribution ?

Have a nice day :)


roles/install_nextcloud/tasks/http_apache.yml
```diff
@@ -47,11 +60,18 @@
   when: (nextcloud_install_tls | bool)
   notify: Restart http
 
+- name: http_apache | Enable Apache2 FPM configuration
```
tty-13, line 65: commit mismatch : this whole block was supposed to go with commit 56d7ab0, sorry.

aalaesar:
Hello there @tty-13 
thank you for your contribution I think that's a good improvement for modularity.
Although this may be a breaking change for current users. could you make it as an optional alternative that can be enabled/disabled with a boolean variable ?

regards,

Aal.

PR #466, opened 2026-07-23
Fix issue with new configuration variable max_memory_limit

pob0111:
PHP 8.5 introduces a new variable named `max_memory_limit`.

The regular expression used to set `memory_limit` also matches that new variable name and thus the wrong setting is overwritten.

By adding `\b` at the beginning and end of the regular expression the full variable name must match.


aalaesar:
Hello @pob0111  thank you for your merge request.

